PR #3, opened 2024-05-02
Add functionality to publish action package

mmokko:
Add functionality to publish action package

New sub commands for package:
  * organizations, list organizations and IDs
  * push, uploads package to Control Room
  * status, gets the package publish status
  * changelog, update the changelog of pushed package
  * publish, publishes the package to Control Room

New command: cloud
  * subcommand: login

Get access credentials from keyring

Read package name from package (.zip) file

Get hostname from keyring

Return JSON output stdout from commands:
  * package organizations
  * package push
  * package status
  * package changelog

What could still be done.
  * Get the hostname from RCC
    * Don't know if RCC has this yet for action server

DO NOT MERGE UNTIL CONTROL ROOM API IS AVAILABLE.


mmokko:
Moved to draft until the: publish API is added.

action_server/src/sema4ai/action_server/_ask_user.py
```diff
@@ -0,0 +1,106 @@
+import typing
+from logging import getLogger
+
+from sema4ai.action_server.package._package_publish import Organization
+
+log = getLogger(__name__)
+
+
+def ask_user_input_to_proceed(msg: str) -> bool:
+    """
```
cmin764, line 11: Might be good to start each docstring with a summary line. (our [convention](https://www.notion.so/robocorptechnologies/Docstrings-c41098ba80cf4ba79036562aba6772b7?pvs=4) for docstrings)

At the end, the linter for them should pass nicely. (Invoke command)

action_server/src/sema4ai/action_server/package/_package_reader.py
```diff
@@ -0,0 +1,54 @@
...
+def read_package_name(package_path: Path) -> str:
+    """
+    Reading package content and extracting the name.
+
+    Arguments:
+        Verified path to built action package (.zip) file
+
+    Returns:
+        Name of the package in the package.yaml
+    """
```
cmin764, line 26:
```suggestion
    """Reading package content and extracting the name.

    Arguments:
        package_path: Verified path to built action package (.zip) file.

    Returns:
        Name of the package in the package.yaml.
    """
```

cmin764 (commented):
Just commented on the docstrings, as the code overall requires testing to see how it behaves first before commenting on it, overall LGTM!

Any timeline for the API availability in order to ship this? (cc @bendersej )

action_server/src/sema4ai/action_server/_ask_user.py
mmokko: This was actually old code, the file was just moved to more generic place. I have followed the convention that was used in this repository. I'm not sure if it's a good idea to start using other convention when adding new docstrings, may be just more confusing.

mmokko:
> Just commented on the docstrings, as the code overall requires testing to see how it behaves first before commenting on it, overall LGTM!
> 
> Any timeline for the API availability in order to ship this? (cc @bendersej )

@cmin764 you can test against the CI. Call the command: cloud login, to set the hostname to: ci.robocorp.dev and it will be saved to the keyring.

action_server/src/sema4ai/action_server/_ask_user.py
```diff
@@ -0,0 +1,106 @@
...
+def ask_user_to_provide_access_credentials() -> str:
+    """
+    Returns:
+        Control Room access credentials in valid form.
+    """
+    from sema4ai.action_server.vendored_deps.termcolors import bold_red
+
+    while True:
+        try:
+            access_credentials = str(input("Enter Control Room access credentials: "))
```
fabioz: Should probably use `getpass.getpass()` here.

fabioz: When we have an `input`, we should handle `KeyboardInterrupt` (Ctrl+C) as a regular part of the flow to cancel (i.e.: without treating it as an error, rather, the user just cancelled it).

It can be done at the caller, but usually I prefer to keep that close to where the input is done and return a `None` and have the caller handle that (so, the return would be `Optional[str]`).

I saw that you're doing it at the caller... so, feel free to keep as is (kept note here just because I had wrote it already ;) ).

action_server/src/sema4ai/action_server/_ask_user.py
```diff
@@ -0,0 +1,106 @@
...
+        Control Room hostname.
+    """
+    from urllib.parse import urlparse
+
+    from sema4ai.action_server._keyring import DEFAULT_CR_HOSTNAME
+    from sema4ai.action_server.vendored_deps.termcolors import bold_red
+
+    while True:
+        try:
+            hostname = input(f"Enter Control Room hostname ({DEFAULT_CR_HOSTNAME}): ")
```
fabioz: Question: instead of `DEFAULT_CR_HOSTNAME`, shouldn't the default be what the user entered the last time (it seems we're saving that, right?)

Or maybe it should show an option with something as default/last/specify new?

action_server/src/sema4ai/action_server/_protocols.py
```diff
@@ -183,3 +191,59 @@ class ArgumentsNamespaceStart(ArgumentsNamespaceBaseImportOrStart):
     expose: bool
     expose_allow_reuse: bool
     api_key: str
+
+
+class ArgumentsNamespacePackageOrganizations(ArgumentsNamespace):
+    command: Literal["package"]
+    package_command: Literal["organizations"]
```
fabioz: Question: does it make sense that getting the organizations is something under `cloud`? Also, just with `organizations` it's a bit unclear whether it's a set or a get. 

Suggestion:

`action-server package list-organizations`

action_server/src/sema4ai/action_server/_protocols.py
```diff
@@ -183,3 +191,59 @@ class ArgumentsNamespaceStart(ArgumentsNamespaceBaseImportOrStart):
...
+    package_command: Literal["state"]
+    access_credentials: Optional[str]
+    organization_id: str
+    package_id: str
+    json: bool
+
+
+class ArgumentsNamespacePackageChangelog(ArgumentsNamespace):
+    command: Literal["package"]
+    package_command: Literal["changelog"]
```
fabioz: Maybe it should be `action-server package set-changelog`? Just by reading it I'm not sure if it's just querying or setting...

As a note, thanks for keeping the styling here (making proper typing is a bit boring, but it does help in knowing what to expect from the command line just from the types). Nice job.

action_server/src/sema4ai/action_server/_protocols.py
```diff
@@ -183,3 +191,59 @@ class ArgumentsNamespaceStart(ArgumentsNamespaceBaseImportOrStart):
...
+
+class ArgumentsNamespaceCloud(ArgumentsNamespace):
+    command: Literal["cloud"]
+    cloud_command: Literal["login"]
+
+
+class ArgumentsNamespaceCloudlogin(ArgumentsNamespace):
+    command: Literal["cloud"]
+    cloud_command: Literal["login"]
+    hostname: str
```
fabioz: Question: I see that `access_credentials` could be used here or could be passed on to each of the other commands explicitly. Should the same be true for the `hostname`?

fabioz (changes requested):
Things missing in the PR:

- Entry in changelog

- Tests

I know the ACE side is still missing, and for testing the upload we need that part working too, but we should at least make sure that the happy path is always working from the action server (in this case it'll also check whether ACE is running -- which I believe is reasonable, if they break it the whole feature breaks).

Note: credentials used should be added to the github secrets. Please coordinate with Kari so that they can be recreated properly when needed (not sure where docs are kept on how to recreate those, but we'll need a way to login to ace with some access credentials for testing purposes.

action_server/src/sema4ai/action_server/_ask_user.py
fabioz: For internal functions, I think it's valid to skip the summary if it'll mostly duplicate what's already in the signature (if it's obvious).

If something isn't obvious, then a summary should be added with that...

Additional note: for things which are part of the public API then it's not the same. All public API functions should have a summary with more details on what the user should expect (and even a change only in the docs could become a breaking change).

action_server/src/sema4ai/action_server/_ask_user.py
mmokko: We are storing it. I think it makes sense to show what the user set last time, if the user wants to just change the access credentials. I'll change it.

action_server/src/sema4ai/action_server/_protocols.py
mmokko: It could be set as an option, but to me it seemed a bit much if user gives the hostname every time, it can be easily forgotten and then the error may be confusing. Although it could make using action-server from CI easier.

I'll add it, makes the API more flexible.

action_server/src/sema4ai/action_server/_protocols.py
mmokko: It's true that there is no package related in this command. I'll move it to be subcommand of cloud command.

action_server/docs/CHANGELOG.md
```diff
@@ -56,8 +69,13 @@
 - Fixed brew reference to `sema4ai/tools/action-server` in conda wrapper.
 - The internal database (which keeps information on the runs, etc) is now auto-migrated, so, the
   separate `migrate` command does not need to be manually called.
+<<<<<<< HEAD
   - Note: The action server will now check the internal database version and if it's newer than
     the current expected version it'll notify the user and abort the startup.
+=======
```
fabioz, line 106: This conflict should keep just the head version.

fabioz (approved):
LGTM (although we can only merge when the other side is working and the tests pass).

action_server/src/sema4ai/action_server/_ask_user.py
cmin764: Fair point, when we'll have that auto-formatter in place @OvidiuCode was thinking about, we'd be able to reformat everything so we'd inline the entire codebase. If the current linting isn't failing, then the format is considered accepted.

PR #8, opened 2024-05-14
action templates handling for new command

el-mike:
Adds Action Template handling for `new` command. 

new `new` argument:
- `template` - allows to specify the name of the template to use. If specified, user won't be prompted to choose one.


action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,110 @@
...
+from pydantic.main import BaseModel
+
+from ._settings import get_default_settings_dir
+
+TEMPLATES_METADATA_URL = "https://downloads.robocorp.com/action-templates/action-templates.yaml"
+TEMPLATES_PACKAGE_URL = "https://downloads.robocorp.com/action-templates/action-templates.zip"
+
+ACTION_TEMPLATES_METADATA_FILENAME = "action-templates.yaml"
+
+ACTION_TEMPLATES_DIR = Path(f"{get_default_settings_dir()}/action-templates")
```
fabioz: This shouldn't be a constant (because `get_default_settings_dir()` may create a directory it should be done as lazily as possible).

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class ActionTemplatesMetadata(BaseModel):
+    hash: str
+    url: str
+    date: datetime.datetime
+    templates: List[ActionTemplate]
+
+
+def _ensure_latest_templates():
```
fabioz: Add return typing (`-> None`).

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,110 @@
...
+        return
+
+    if not local_metadata or local_metadata.hash != new_metadata.hash:
+        _download_and_unzip_templates(ACTION_TEMPLATES_DIR)
+
+        with open(ACTION_TEMPLATES_METADATA_PATH, "w+") as f:
+            f.write(new_metadata_content)
+
+
+def _download_and_unzip_templates(action_templates_dir: Path):
```
fabioz: Add return typing.

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,110 @@
...
+def _get_local_templates_metadata() -> Optional[ActionTemplatesMetadata]:
+    # Loads templates metadata YAML file.
+
+    if not os.path.isfile(ACTION_TEMPLATES_METADATA_PATH):
+        return None
+
+    return _parse_templates_metadata(ACTION_TEMPLATES_METADATA_PATH.read_text())
+
+def _parse_templates_metadata(yaml_content: str) -> Optional[ActionTemplatesMetadata]:
+    # Parses templates YAML metadata into ActionTemplatesMetadata model.
```
fabioz: I think the first line comments should be docstrings, not just comments (although if it'll just say the same thing as the function signature, feel free to skip it).

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,110 @@
...
+
+from ._settings import get_default_settings_dir
+
+TEMPLATES_METADATA_URL = "https://downloads.robocorp.com/action-templates/action-templates.yaml"
+TEMPLATES_PACKAGE_URL = "https://downloads.robocorp.com/action-templates/action-templates.zip"
+
+ACTION_TEMPLATES_METADATA_FILENAME = "action-templates.yaml"
+
+ACTION_TEMPLATES_DIR = Path(f"{get_default_settings_dir()}/action-templates")
+ACTION_TEMPLATES_METADATA_PATH = Path(f"{ACTION_TEMPLATES_DIR}/{ACTION_TEMPLATES_METADATA_FILENAME}")
```
fabioz: Likewise, as this depends on the `ACTION_TEMPLATES_DIR` it must be changed to be a function.

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,110 @@
...
+    # It downloads the latest templates metadata file, and compares the hash with
+    # the metadata hold locally (if exists).
+    # If there is no match (or metadata is not available locally), it will download
+    # the templates package.
+
+    os.makedirs(ACTION_TEMPLATES_DIR, exist_ok=True)
+
+    local_metadata = _get_local_templates_metadata()
+
+    new_metadata_content = requests.get(TEMPLATES_METADATA_URL).text
```
fabioz: As accessing the network is common to fail, this should probably show a nicer message when it fails -- also, if templates were already in place, maybe it can just proceed with a log message and use what's already downloaded.

fabioz (commented):
The major thing missing are the tests.

The action serve is very much test-driven in the codebase, so, code should have tests when added (at the minimum the happy path should be tested to make sure it works and possibly some cases such as when the network is not available but templates are already there).

We're not aiming for 100% test coverage, but at least the regular expected paths should be tested (personally I usually work test-first and doing it step by step as I find that doing it later has the psychological effect that "things are already working, why bother to do it now?", whereas doing it first enables it to be a regular part of development and -- at least for me -- makes the code implementation be faster than doing tests afterwards as it becomes a part of the development flow).

fabioz:
Also, there should be an entry in the `changelog.md` for each PR with the changes that were done.

action_server/src/sema4ai/action_server/_new_project_helpers.py
el-mike: I was following the guidelines we have on documenting private member:
https://www.notion.so/robocorptechnologies/Docstrings-c41098ba80cf4ba79036562aba6772b7?pvs=4#8cfd56c066ca4a31af928d50c290f87d

Happy to change it if we want to use docstring for private bits as well ;)

action_server/src/sema4ai/action_server/_new_project_helpers.py
el-mike: I decided to leave error handling to the caller (`_new_project.py` -> `create_new_project`) - let me know if you think that's okay, or we should add more logs closer to the place the error is raised.
In any case, if `_ensure_latest_templates` fails at any point, a message will be shown to the user, but the main command will continue, using already cached templates (if available).

action_server/src/sema4ai/action_server/_new_project_helpers.py
el-mike: done

action_server/src/sema4ai/action_server/_new_project_helpers.py
el-mike: done

action_server/src/sema4ai/action_server/_new_project_helpers.py
el-mike: done

action_server/src/sema4ai/action_server/_new_project_helpers.py
el-mike: I've removed some of the rather redundant comments ; )

action_server/src/sema4ai/action_server/_new_project_helpers.py
el-mike: done

action_server/docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
   from the action server notifying users about it if the old directory exists).
 - The `SEMA4AI_HOME` environment variable can now be used to set a different home
   folder to store the `action-server` directory information.
+- `new` command allows to choose a template from a predefined list now. Command will list available templates,
+  and prompt user to choose one. 
+  - `--template` argument has been, allowing to specify a concrete template when running the command (when specified,
```
fabioz: It feels like something is missing in this phrase...
`--template` argument has been, allowing`

action_server/src/sema4ai/action_server/_new_project.py
```diff
@@ -1,19 +1,37 @@
...
 
 log = logging.getLogger(__name__)
 
 
-def create_new_project(directory: str = "."):
+def create_new_project(directory: str = ".", template_name: str = "") -> None:
+    """Creates a new project under the specified directory.
+
+    Args:
+        directory (str): The directory to create the project in.
```
fabioz: No need to add the `(str)` here as the type is in the signature.

action_server/src/sema4ai/action_server/_new_project.py
```diff
@@ -1,19 +1,37 @@
...
+        template_name (str): Temple to use for the new project.
```
fabioz: typo `Temple` -> `Template`.

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+
+    if not os.path.isfile(template_path):
+        raise RuntimeError(f"Template {template_name} does not exist")
+
+    with zipfile.ZipFile(template_path, "r") as zip_ref:
+        zip_ref.extractall(directory)
+
+
+def _get_action_templates_dir_path() -> Path:
+    return Path(f"{get_default_settings_dir()}/action-templates")
```
fabioz: This construct is a bit weird. Should probably be:

`get_default_settings_dir() / "action-templates"`

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def _get_action_templates_metadata_path() -> Path:
+    return Path(f"{_get_action_templates_dir_path()}/{ACTION_TEMPLATES_METADATA_FILENAME}")
```
fabioz: Should probably be:

`_get_action_templates_dir_path() / ACTION_TEMPLATES_METADATA_FILENAME`

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+            date=metadata.get("date", None),
+            templates=templates
+        )
+    except yaml.YAMLError as e:
+        log.warning(f"Error reading metadata: {e}")
+        return None
+
+
+def _unpack_template(template_name: str, directory: str = ".") -> None:
+    template_path = f"{_get_action_templates_dir_path()}/{template_name}.zip"
```
fabioz: I think it could use just the `pathlib.Path` api.

i.e.:

```
template_path = _get_action_templates_dir_path() / f"{template_name}.zip"

if not template_path.is_file():
    ...
```

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+        return None
+
+    return _parse_templates_metadata(action_templates_metadata_path.read_text())
+
+
+def _parse_templates_metadata(yaml_content: str) -> Optional[ActionTemplatesMetadata]:
+    try:
+        metadata: ActionTemplatesYaml = yaml.safe_load(yaml_content)
+
+        templates: List[ActionTemplate] = list()
```
fabioz: Using `list()` is a bit unpythonic. Better would be just `[]`.

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+
+        for name, description in metadata.get("templates", dict()).items():
```
fabioz: Using `dict()` is a bit unpythonic. Better would be just `{}`.

action_server/src/sema4ai/action_server/_new_project_helpers.py
fabioz: As action server is only Python 3.10 onwards, we can use `list[ActionTemplate]` (instead of `List[ActionTemplate]`).

action_server/src/sema4ai/action_server/_new_project_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+ACTION_TEMPLATES_METADATA_FILENAME = "action-templates.yaml"
+
+log = logging.getLogger(__name__)
+
+
+class ActionTemplatesYaml(TypedDict):
+    hash: str
+    url: str
+    date: datetime.datetime
+    templates: Dict[str, str]
```
fabioz: We can use 'dict[str, str]` here as the action server is Python 3.10 onwards.

action_server/src/sema4ai/action_server/_new_project.py
```diff
@@ -1,19 +1,38 @@
...
-
-TEMPLATE_URL = (
-    "https://github.com/robocorp/template-action/archive/refs/heads/master.zip"
-)
 
 log = logging.getLogger(__name__)
 
 
-def create_new_project(directory: str = "."):
+def create_new_project(directory: str = ".", template_name: str = "") -> None:
```
fabioz: I was actually taking a look in the code and this should be changed to return 0 or an error code (1 is fine) and the `_main_retcode` should return the error code in that case too.

Note that it's fine to return 0 if it was cancelled with a `Ctrl+C`, but if an exception happens we should print it and return a non-0 value.

el-mike: @fabioz done!

fabioz (approved):
LGTM!

cmin764 (approved):
Quickly skimmed through the changes and LGTM! (I see Fabio already made a thorough review and he has the best context here)

action_server/docs/CHANGELOG.md
```diff
@@ -2,14 +2,14 @@
...
-
-- Accept passing `OAuth2 secrets` (as a dict with the required keys) in the `/api/secrets`.
 - `new` command allows to choose a template from a predefined list now. Command will list available templates,
   and prompt user to choose one. 
   - `--template` argument has been added, allowing to specify a concrete template when running the command (when specified,
     user won't be prompted to choose a template from the list).
 
+## 0.9.1 - 2024-05-17
+
+- Accept passing `OAuth2 secrets` (as a dict with the required keys) in the `/api/secrets`.
```
cmin764: I was looking for this entry [here](https://github.com/Sema4AI/actions/blob/master/actions/docs/CHANGELOG.md) but I couldn't find it. What happened?

fabioz: You're looking into the changelog for actions... the proper one is the one from the action server: https://github.com/Sema4AI/actions/blob/master/action_server/docs/CHANGELOG.md

PR #12, opened 2024-05-20
action templates

el-mike:
Adds Action Templates to the repository, alongside workflows and tooling for deployment.

NOTE:
Reviews of README files would be appreciated (to double check whether they fall into correct standards)


fabioz (approved):
I think templates themselves can be improved (for instance, on the advanced one I think an action package with an actual python package with `__init__.py` and more than one action separated in `*action*.py` could be added to showcase that you're not constrained by the `action.py` structure).

Still, I think this should be done afterwards as this PR is more about getting the scaffolding in place, so, LGTM :)

el-mike:
@fabioz yeah, I was thinking about something like that - we decided however, that we want to keep the size of the templates minimal, so the users don't need to delete too much stuff when scaffolding a project with it. @kariharju your input would be appreciated here ; ))

PR #13, opened 2024-05-21
list-templates subcommand

el-mike:
Adds "list-templates" subcommand for the "new" command, allowing to get templates available for a new project.

Subcommand accepts "--json" flag - when set, output will be formatted as JSON. 

NOTE:
Tests currently check against stderr, as it seems like `new` command writes its output there. Further investigation needed

### DIFF
--- a/action_server/src/sema4ai/action_server/_new_project.py
+++ b/action_server/src/sema4ai/action_server/_new_project.py
@@ -1,5 +1,13 @@
+import json
 import logging
 import os
+import typing
+
+from ._protocols import (
+    ArgumentsNamespace,
+    ArgumentsNamespaceNew,
+    ArgumentsNamespaceNewTemplates,
+)
 
 log = logging.getLogger(__name__)
 
@@ -11,15 +19,12 @@ def create_new_project(directory: str = ".", template_name: str = "") -> int:
         directory: The directory to create the project in.
         template_name: Template to use for the new project.
     """
-    from sema4ai.action_server.vendored_deps.termcolors import (
-        bold_red,
-        bold_yellow,
-        colored,
-    )
+    from sema4ai.action_server.vendored_deps.termcolors import bold_red, bold_yellow
 
     from ._new_project_helpers import (
         _ensure_latest_templates,
         _get_local_templates_metadata,
+        _print_templates_list,
         _unpack_template,
     )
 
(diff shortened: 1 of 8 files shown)


action_server/src/sema4ai/action_server/_new_project.py
```diff
@@ -83,3 +87,51 @@ def create_new_project(directory: str = ".", template_name: str = "") -> int:
         )
 
         return 1
+
+
+def list_templates(print_json: bool = False) -> int:
```
fabioz: Suggestion: rename to `_handle_list_templates` (as `list_templates` seems like it'll return the templates, instead of printing them).

action_server/src/sema4ai/action_server/_new_project.py
```diff
@@ -83,3 +87,51 @@ def create_new_project(directory: str = ".", template_name: str = "") -> int:
...
+        _ensure_latest_templates,
+        _get_local_templates_metadata,
+        _print_templates_list,
+    )
+
+    try:
+        _ensure_latest_templates()
+        metadata = _get_local_templates_metadata()
+
+        if not metadata or len(metadata.templates) == 0:
```
fabioz: Humm, not sure it should raise an error here. For this function printing `[]` if json or printing `No templates available` if not json seems fine.

action_server/src/sema4ai/action_server/_new_project.py
el-mike: Yeah, I was considering that - however, it feels like `create_new_project` would fall a bit out of convention for this file. Do you reckon changing it to `handle_new_project` would make sense?

action_server/src/sema4ai/action_server/_new_project.py
el-mike: I was also on the fence - decided however, that because no templates would make `new` command itself raise an error, decided to do the same here. Let me know what you think !

action_server/src/sema4ai/action_server/_new_project.py
fabioz: I agree.

action_server/src/sema4ai/action_server/_new_project.py
fabioz: I think it makes sense for new because it cannot create a project without a template, but for listing an empty list is something valid.

action_server/src/sema4ai/action_server/_new_project.py
el-mike: Great! I think it will also align with Matti's changes nicely, as he used similar naming convention

action_server/src/sema4ai/action_server/_new_project.py
el-mike: Fair - I will update it!

fabioz (approved):

PR #15, opened 2024-05-22
Feat/response

fabioz:
Although this is in draft, the main idea should be there already, so, checking for feedback on what you think about it @cmin764 @vlad-sema4


action_server/src/sema4ai/action_server/_actions_run.py
```diff
@@ -222,34 +230,49 @@ def _run_action_in_thread(
...
+                    raise RuntimeError(
+                        "It was not possible to collect the contents of the "
+                        "result (json not created)."
+                    )
+
+                try:
+                    result_contents = json.loads(run_result_str)
+                except Exception:
+                    raise RuntimeError(
+                        f"Error loading the contents of {run_result_str} as json."
```
cmin764, line 245:
```suggestion
                        f"Error loading the contents of {run_result_str!r} as json."
```
If you want it repr-ed.

action_server/src/sema4ai/action_server/_actions_run.py
```diff
@@ -222,34 +230,49 @@ def _run_action_in_thread(
...
+                    )
+
+                ret = result_contents.get("result")
+                result_str: str
+                if ret is None:
+                    result_str = "None"
```
cmin764:
```suggestion
                    result_str = "null"
```
if we'd like to inline it with JSON serialized value of Python's `None`. (but your preference here)

action_server/src/sema4ai/action_server/_actions_run.py
```diff
@@ -222,34 +230,49 @@ def _run_action_in_thread(
...
+                else:
+                    result_str = json.dumps(ret, indent=4)
```
cmin764:
```suggestion
                result_str = json.dumps(result_contents.get("result"), indent=4)
```
isn't this simpler and more clean?

Example:
```python
>>> json.dumps({}.get("x"))
'null'
```

actions/pyproject.toml
```diff
@@ -17,14 +17,12 @@ robocorp-log = ">=2.4,<3"
 robocorp-truststore= ">=0.8.0"
 psutil = "^5.0"
 docstring_parser_fork = "^0.0.5"
+pydantic = "^2.0" # Required for the `Response[T]`
```
cmin764, line 20: Left a Slack [thread](https://sema4ai.slack.com/archives/C06CYLQ7S4R/p1716448556918039?thread_ts=1716382901.166559&cid=C06CYLQ7S4R).

I might be off-topic, just an idea.

actions/src/sema4ai/actions/_commands.py
```diff
@@ -536,6 +579,14 @@ def run(
...
                 else:
                     retcode = int(any(action.failed for action in actions))
+                    # retcode = 0
+                    # for action in actions:
+                    #     if action.failed:
+                    #         if action.result is None:
+                    #             # This is the case where an ActionError is raised
+                    #             # (so, we have an exception but also a proper result).
+                    #             retcode = 1
+                    #             break
```
cmin764: Does it help to have dead code here? (or is this more like a reference of a potential upcoming TODO)

cmin764 (approved):
LGTM, left mostly optional nitpicks.

actions/src/sema4ai/actions/_commands.py
fabioz: No, mostly leftovers from some experimentation. I'll definitely remove before the final version.

action_server/src/sema4ai/action_server/_actions_run.py
fabioz: I was thinking about that too (it breaks tests, which is why I haven't done it, but I may end up doing that, just need to think a bit more about it).

action_server/src/sema4ai/action_server/_actions_run.py
fabioz: Given that it's an indented json I think it's better not to do that (so messages may be more well formatted to show in the terminal).

action_server/src/sema4ai/action_server/_actions_run.py
fabioz: Changed (but with `None` still being added to the error message)

PR #23, opened 2024-05-31
Fix keyring load backend issue

mmokko:
Tested on windows and it can load the backend, though would still need to test the bundled package.
Testing on macos and it can load the backend, of course need to test this with the bundled package as well.


cosmogp (approved):
LGTM

PR #24, opened 2024-06-04
Matti/remove keyring

mmokko:
Stores the data needed to be stored for consecutive runs to a file: ~/.sema4ai/action_server/action_server_storage.json

It currently saves { cloud: { access_credentials: { iv: str, tag, str, value, str }, hostname: str }}

The encryption is done using the encrypt API in place.
We could hardcode the key into the codebase and then overwrite the value in github actions build pipeline. Though even that can be extracted quite easily using hexdump.

The key is now written to a file in a disk next to the storage file, so at least it has encryption at rest.


cosmogp (approved):
LGTM

action_server/src/sema4ai/action_server/_storage.py
```diff
@@ -0,0 +1,159 @@
...
+from pathlib import Path
+
+from pydantic.main import BaseModel
+
+from sema4ai.action_server._settings import get_default_settings_dir
+
+log = getLogger(__name__)
+
+STORAGE_FILE_NAME = "action_server_storage.json"
+STORAGE_PATH = Path(os.path.join(get_default_settings_dir(), STORAGE_FILE_NAME))
```
fabioz: `get_default_settings_dir` should not be called at top level (it has a side-effect of creating a directory). This means that instead of `STORAGE_PATH` being a constant it should be a function such as `get_storage_path`.

action_server/src/sema4ai/action_server/_storage.py
```diff
@@ -0,0 +1,159 @@
...
+    hostname: str
+    access_credentials: StorageEncryptedValue
+
+
+class Storage(BaseModel):
+    cloud: StorageCloud
+
+
+@lru_cache
+def get_key() -> bytes:
```
fabioz, line 53: Humm, I'm not sure I get it... if we're storing the key (in which case if the system gets compromised they'll get the whole datadir anyways), what's the point of the encryption?

If we choose to drop the keyring then this encryption doesn't really do much sense because all of the info is compromised anyways.

fabioz: As we talked, the key should be saved inside the user dir with strict permissions.

mmokko: Updated to have a static sema4ai path that always points into users folder.

action_server/src/sema4ai/action_server/_storage.py
```diff
@@ -0,0 +1,174 @@
...
+    key_file_path = get_key_file_path()
+    if not key_file_path.exists():
+        key = os.urandom(AES_KEY_LEN)
+        with open(key_file_path, "wb") as f:
+            f.write(key)
+    else:
+        with open(key_file_path, "rb") as f:
+            key = f.read()
```
fabioz: I think the key should be validated to check whether it's what you expect (i.e.: it has the proper size). If it's not the fallback should be to re-create it.

action_server/src/sema4ai/action_server/_storage.py
```diff
@@ -0,0 +1,174 @@
...
+class Storage(BaseModel):
+    cloud: StorageCloud
+
+
+@lru_cache
+def get_key() -> bytes:
+    key_file_path = get_key_file_path()
+    if not key_file_path.exists():
+        key = os.urandom(AES_KEY_LEN)
+        with open(key_file_path, "wb") as f:
```
fabioz: The proper access permissions should be set in the file.

fabioz (approved):

PR #25, opened 2024-06-06
From package build print json output for stdout

mmokko:
Print the pat to the built package to the stdout in json format if argument --json is added.

VSCode extension needs to locate the package once it's built.

### DIFF
--- a/action_server/src/sema4ai/action_server/package/_package_build_cli.py
+++ b/action_server/src/sema4ai/action_server/package/_package_build_cli.py
@@ -1,3 +1,4 @@
+import json
 import os
 import sys
 import time
@@ -144,6 +145,7 @@ def add_package_command(command_subparser, defaults):
     )
     add_data_args(build_parser, defaults)
     add_verbose_args(build_parser, defaults)
+    add_json_output_args(build_parser)
 
     ### Extract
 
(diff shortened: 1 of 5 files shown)


action_server/src/sema4ai/action_server/package/_package_build_cli.py
```diff
@@ -275,20 +277,23 @@ def handle_package_command(base_args: ArgumentsNamespace):
...
         except ActionServerValidationError as e:
             log.critical(
                 bold_red(
                     f"\nUnable to build action package. Please fix the error below and retry.\n{e}",
                 )
             )
-            retcode = 1
-        return retcode
+            return 1
+        return result.return_code
```
cosmogp, line 296: The function might need some clear return types.

cosmogp (approved):

fabioz (approved):
LGTM

action_server/src/sema4ai/action_server/package/_package_build_cli.py
mmokko: Would be nice, the whole file basically uses return code 1 for all errors, so I wont do it here. It would be a bigger refactoring.

PR #28, opened 2024-06-06
Template updates to package.yamls

kariharju:
Changed documentation to point to relative file.
Added version -field back

### DIFF
--- a/templates/advanced/package.yaml
+++ b/templates/advanced/package.yaml
@@ -4,9 +4,6 @@ name: Advanced Template Package
 # Required: A description of what's in the action package.
 description: Action package description
 
-# Required: Point to main documentation markdown file for the package.
-documentation: README.md
-
 # Package version number, recommend using semver.org
 version: 0.0.1
 
--- a/templates/advanced/package.yaml
+++ b/templates/advanced/package.yaml
@@ -4,15 +4,18 @@ name: Advanced Template Package
 # Required: A description of what's in the action package.
 description: Action package description
 
-# Required: A link to where the documentation on the package lives.
-documentation: https://github.com/...
+# Required: Point to main documentation markdown file for the package.
+documentation: README.md
+
+# Package version number, recommend using semver.org
+version: 0.0.1
 
 dependencies:
   conda-forge:
   - python=3.10.14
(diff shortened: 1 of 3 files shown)


templates/advanced/package.yaml
```diff
@@ -4,15 +4,18 @@ name: Advanced Template Package
 # Required: A description of what's in the action package.
 description: Action package description
 
-# Required: A link to where the documentation on the package lives.
-documentation: https://github.com/...
+# Required: Point to main documentation markdown file for the package.
+documentation: README.md
```
cmin764: Have we decided on picking configuration over convention? Like, what would be an use-case where the `documentation:` field won't be a _README.md_? Normally, any action pack contains the main _README.md_, which can also point to additional documentation found in the _docs_ dir if needed.

The _package.png_ idiom works great so far, so what would be an advantage here on bringing them in the _package.yaml_?

templates/advanced/package.yaml
```diff
@@ -4,15 +4,18 @@ name: Advanced Template Package
...
+
+# Package version number, recommend using semver.org
+version: 0.0.1
 
 dependencies:
   conda-forge:
   - python=3.10.14
   - uv=0.2.4
```
cmin764, line 13: I saw some variations of `uv=0.2.6` here, while the latest is `uv=0.2.9`. What's the rationale when picking the right version and why would **0.2.4** be better than the latest one which is **0.2.9**?

cmin764 (commented):
Placed two questions that should be answered clearly before on proceeding updating those, so we'd learn on how to act about it from now on.

templates/advanced/package.yaml
kariharju: Good point, we agreed to convention on this.

templates/advanced/package.yaml
kariharju: Bi-weekly update cadence needs to be set.. so not everyone updating to the latest "whenever" but a set time when we update and test all to a specific uv version.
We cannot have 12 different environment contents, we rather need more of a "golden env." that all of these try to use.

el-mike (approved):

PR #67, opened 2024-06-17
Use network settings from RCC profile

mmokko:
Read the RCC profile settings and take them into use when doing http requests.

Create a singleton requests.Session object that has the settings enabled.
Use the singleton instead of requests.

TODO: should we return from action server if we fail to read the RCC profile settings? Currently it will just log a critical message but will proceed.

### DIFF
--- a/action_server/src/sema4ai/action_server/_cli_impl.py
+++ b/action_server/src/sema4ai/action_server/_cli_impl.py
@@ -553,54 +553,63 @@ def _main_retcode(
         download_rcc(target=download_args.file, force=True)
         return 0
 
-    if command == "package":
-        from sema4ai.action_server.package._package_build_cli import (
-            handle_package_command,
+    from sema4ai.action_server._rcc import initialize_rcc
+    from sema4ai.action_server._download_rcc import download_rcc
+    from sema4ai.action_server._session import initialize_session
+
+    with initialize_rcc(download_rcc(), None) as rcc:
+        initialize_session(rcc)
+
+        if command == "package":
+            from sema4ai.action_server.package._package_build_cli import (
+                handle_package_command,
+            )
+
+            return handle_package_command(base_args)
+
+        if command == "env":
+            from sema4ai.action_server.env import handle_env_command
+
+            return handle_env_command(base_args)
+
+        if command == "cloud":
+            from sema4ai.action_server._actions_cloud import handle_cloud_command
+
+            return handle_cloud_command(base_args)
+
+        if command not in (
+            "migrate",
+            "import",
+            "start",
+            "new",
+            "cloud",
+        ):
+            log.critical(f"Unexpected command: {command}.")
+            return 1
+
+        log.info(
+            colored("\n  ⚡️ Starting Action Server... ", attrs=["bold"])
+            + colored(f"v{__version__}\n", attrs=["dark"])
         )
 
-        return handle_package_command(base_args)
+        if command == "new":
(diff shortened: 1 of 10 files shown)


action_server/src/sema4ai/action_server/_session.py
```diff
@@ -0,0 +1,84 @@
...
+
+if TYPE_CHECKING:
+    from sema4ai.action_server._rcc import Rcc
+
+
+log = logging.getLogger(__name__)
+
+
+class RccCertificateSettings(BaseModel):
+    verify_ssl: Optional[bool] = Field(default=None, alias="verify-ssl")
```
kariharju, line 16: Is there yet a way to set the legacy_renegotiation via the options?

templates/advanced/actions.py
```diff
@@ -33,20 +30,23 @@ def get_repository_commits(
     Returns:
         A list of commit messages.
     """
+    from sema4ai.action_server._session import session
```
fabioz: I don't think this can be done as is... the `templates` can't really import the `action_server`. If we need the user to use rcc settings in actions we'd need a different design.

Wouldn't it be enough for actions to have the env variables set by the action server?

action_server/src/sema4ai/action_server/_cli_impl.py
```diff
@@ -553,54 +553,63 @@ def _main_retcode(
...
-        return handle_env_command(base_args)
+            return handle_package_command(base_args)
 
-    if command == "cloud":
-        from sema4ai.action_server._actions_cloud import handle_cloud_command
+        if command == "env":
+            from sema4ai.action_server.env import handle_env_command
 
-        return handle_cloud_command(base_args)
+            return handle_env_command(base_args)
```
fabioz: Since `rcc` is already initialized here, I guess it should be passed to the `handle_env_command` now (maybe `initialize_rcc` should be changed to check if `_rcc` is not `None` and if it is raise an error to avoid initializing multiple times).

action_server/src/sema4ai/action_server/_session.py
mmokko: Added support for this. The session context has the option set if the underlying SSL library supports it. There is a log line if it's not supported.

We cannot propagate this to the actions though as there is no environment variable for this.

templates/advanced/actions.py
mmokko: Thanks for noticing this 👍 I didn't realize this was another context.

Yes we can just set the environment variables as we are. Unfortunately the SSL verify option doesn't seem to work with just the environment variable. Also the allow legacy renegotiation doesn't have any environment variable.

action_server/src/sema4ai/action_server/_cli_impl.py
mmokko: Very good point 👍 did the changes.

action_server/src/sema4ai/action_server/_rcc.py
```diff
@@ -356,12 +360,16 @@ def clean_tools_caches(self):
...
     global _rcc
 
-    rcc = Rcc(rcc_location, robocorp_home)
-    _rcc = rcc
-    try:
-        yield rcc
-    finally:
-        _rcc = None
+    if _rcc:
+        yield _rcc
```
fabioz, line 364: Minor: should have a return after this yield so that it doesn't even go to the other check.

fabioz (approved):

PR #88, opened 2024-08-12
Updating product's home directory strategy

el-mike:
- Bumps Action Server's RCC version to latest (v18.1.5)
- Adds `--sema4ai` flag to RCC calls of AS is run with the `sema4ai` strategy
- updates Go wrapper to use `sema4ai` folders
- Includes some minor renames
- Includes some CDN URL updates
- Includes updates to test suite to use new naming

### DIFF
--- a/action_server/go-wrapper/main.go
+++ b/action_server/go-wrapper/main.go
@@ -10,7 +10,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"slices"
 	"strings"
 
 	"golang.org/x/mod/semver"
@@ -134,14 +133,6 @@ func main() {
 	var actionServerPath string
 	var executablePath string
 
-	var baseSettingsDirName string
-
-	if slices.Contains(os.Args, "--sema4ai") {
-		baseSettingsDirName = "sema4ai"
-	} else {
-		baseSettingsDirName = "robocorp"
-	}
-
 	// Read the version
 	versionData, err := content.ReadFile("assets/version.txt")
 	if err != nil {
@@ -157,7 +148,7 @@ func main() {
 	switch runtime.GOOS {
 	case "windows":
 		appDataDir := os.Getenv("LOCALAPPDATA")
-		actionServerPath = fmt.Sprintf("%s\\%s\\action-server\\%s", appDataDir, baseSettingsDirName, version)
+		actionServerPath = fmt.Sprintf("%s\\sema4ai\\action-server\\%s", appDataDir, version)
 		executablePath = filepath.Join(actionServerPath, "action-server.exe")
 	case "linux", "darwin":
 		homeDir, err := os.UserHomeDir()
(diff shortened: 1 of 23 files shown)


action_server/src/sema4ai/action_server/_settings.py
```diff
@@ -42,15 +42,15 @@ def get_python_exe_from_env(env):
...
+def _legacy_get_default_settings_dir() -> Path:
     if sys.platform == "win32":
         localappdata = os.environ.get("LOCALAPPDATA")
         if not localappdata:
             raise RuntimeError("Error. LOCALAPPDATA not defined in environment!")
         path = Path(localappdata) / "robocorp" / ".action_server"
     else:
         # Linux/Mac
-        path = Path("~/robocorp/.action_server").expanduser()
+        path = Path("~/.robocorp/.action_server").expanduser()
```
el-mike, line 53: @fabioz please verify if we want to update it - I recall talking about missing the `.` was a mistake

action_server/go-wrapper/main.go
```diff
@@ -133,6 +134,14 @@ func main() {
 	var actionServerPath string
 	var executablePath string
 
+	var baseSettingsDirName string
+
+	if slices.Contains(os.Args, "--sema4ai") {
```
fabioz: For the go wrapper, we don't need this. It'll always be `--sema4ai` (so, all the code related to using `robocorp` folders can be removed).

action_server/src/sema4ai/action_server/_cli_impl.py
```diff
@@ -52,6 +52,16 @@ def _add_whitelist_args(parser, defaults):
     )
 
 
+def _add_dir_strategy_args(parser, defaults):
```
fabioz: Same thing here. This should not be an option, only using `sema4ai` folders is supported in the action server.

fabioz (changes requested):

action_server/go-wrapper/main.go
el-mike: @fabioz done

action_server/src/sema4ai/action_server/_cli_impl.py
el-mike: @fabioz done

action_server/src/sema4ai/action_server/_settings.py
cosmogp: It is. 

Actually, this might need better handling. 
Not for this PR but there is a good example in Studio how we calculate the path cross-platform. https://github.com/Sema4AI/sema4ai-studio/blob/43e9ba37ce5170b42db59430fa6e92d699fb267a/src/common/pathHandler.ts

cosmogp (approved):
LGTM

fabioz (approved):
LGTM

PR #92, opened 2024-08-22
exposing sema4.ai default OAuth2 config

el-mike:
Adds Sema4.ai's default OAuth config (taken from private repository) to Action Server static files, that are available as source code files after build, similarly to frontend `index.html` file.

Since AS is distributed either via pip or binary, the config had to be available through both channels - therefore, embedding it in Go wrapper couldn't be a solution.

Since the config lives in Python file after build, clients have to ask Action Server for it explicitly, using `action-server get-default-oauth-config` command. Note that the command is not added to the general help contents of the executable, as it meant to be used internally.

No entry to `CHANGELOG.md` was added, as the change is meant to be used internally only.


kariharju (approved):

.github/workflows/action_server_manylinux_release.yml
```diff
@@ -5,16 +5,15 @@ name: Action Server [MANYLINUX] Release
 
 on:
   push:
-#     branches:
-#       - "wip-manylinux"
+    #     branches:
```
mmokko, line 8: The indentation looks off here, should there be 2 spaces instead of 4?

action_server/src/sema4ai/action_server/_cli_impl.py
```diff
@@ -289,6 +289,14 @@ def _add_new_command(command_subparser, defaults):
 
     add_verbose_args(list_templates_parser, defaults)
     add_json_output_args(list_templates_parser)
+    
+    
+def _add_get_default_oauth_config_command(command_subparser, defaults):
+    # This serves only as an internal command, therefore we don't add help entry to it.
+    command_subparser.add_parser(
+        "get-default-oauth-config",
```
mmokko: I guess this is the default sema4.ai config? Should the name be: get-sema4ai-oauth-config?

mmokko (approved):
Hey, it looks good to me 👍 

I added one comment about renaming the command.

.github/workflows/action_server_manylinux_release.yml
el-mike: I think it was Kari's autoformat - it does have 2 indents now though, it's only 4 after the comment ^^ but the comment itself starts at indent 2

action_server/tests/action_server_tests/test_cli/test_help.txt
```diff
@@ -1,11 +1,11 @@
 usage: action-server [-h]
-                     {start,import,download-rcc,new,version,migrate,package,env,cloud}
+                     {start,import,download-rcc,new,version,migrate,package,env,cloud,get-sema4ai-oauth-config}
                      ...
 
 Sema4.ai Action Server (<version>)
 
 positional arguments:
-  {start,import,download-rcc,new,version,migrate,package,env,cloud}
+  {start,import,download-rcc,new,version,migrate,package,env,cloud,get-sema4ai-oauth-config}
```
fabioz: This change means that it's actually visible to the user... To make it really invisible I guess you have to check `sys.args` directly before giving it to the arguments parser instead of doing it through the arguments parser if it's really meant to be invisible to the user.

-- If it can be shown to the user that'd be ok.

kariharju: I think this is ok, in any case the Action Server cli needs to give these out. We also have changes in mind for this already. So I'd get this version in.

el-mike: @fabioz @kariharju I did remove it ; ))

PR #94, opened 2024-08-27
adding get-user-oauth-config-path command

el-mike:
- Add top level command: `oauth2`, providing OAuth2 related utilities
- Add `sema4ai-config` subcommand for `oauth2`
  - Returns OAuth2 configuration for Sema4.ai provided OAuth2 applications 
- Add `user-config-path` subcommand for `oauth2`
  - Returns the path to user's local OAuth config file
  - `--json` argument can be provided to get the result in JSON format

- Renames some references from `oauth` to `oauth2` in file names, values and variables.

The default user's config is saved in `_oauth2_config.py` during build time, similarly to Sema4.ai config.


mmokko (approved):
Looks good to me 👍

action_server/tests/action_server_tests/test_cli/test_help.txt
```diff
@@ -15,6 +15,8 @@ positional arguments:
     package             Utilities to manage the action package
     env                 Commands related to the environment handling
     cloud               Utilities to perform Control Room operations
+    get-user-oauth-config-path
```
fabioz: I think the command structure could have an `oauth2`  with subcommands (I think it may grow, so, preparing for that...).

So, it can be something as:

`action-server oauth2 user-config-path`

Note all references should be to `oauth2` instead of `oauth`.

fabioz (changes requested):

action_server/tests/action_server_tests/test_cli/test_help.txt
el-mike: @fabioz done!

PR #105, opened 2024-09-23
Fixed issue with download with make_executable=True, fixed typing issues, made as package to mark as `py.typed`

fabioz:
Fixed issues with download, typing, package structure.

Also:
- Fixed issue with download with make_executable=True
- Fixed typing issues
- Made as package to mark as `py.typed`
- Changed tasks.py to use the same structure of other packages
- Using SEMA4AI_TLS_LEGACY_RENEGOTIATION_ALLOWED or RC_TLS_LEGACY_RENEGOTIATION_ALLOWED to define whether to enable legacy ssl connect.
- Added tests


fabioz:
I need to use it in the new files API, so, fixed some issues to make a new release (the most annoying was that although it was typed the python typing didn't pick it up because `py.typed` wasn't set -- that's the reason I had to convert it to a package too).

PR #121, opened 2024-11-01
Feat/files

mkorpela:
So far tested against locally running Sema4.ai Studio backend.


actions/src/sema4ai/actions/chat/_client.py
```diff
@@ -156,65 +153,69 @@ def set_bytes(
...
             file_ref = response_data["file_ref"]
 
-            parsed_url = urllib.parse.urlparse(file_url)
-            if parsed_url.scheme == "file":
-                # We need to add the file to the local file system
-                p = Path(_uris.to_fs_path(file_url))
-                p.parent.mkdir(parents=True, exist_ok=True)
-                p.write_bytes(content)
+            # If the URL is a local path, write directly to filesystem
+            if upload_url.startswith('/'):
```
fabioz: The old thing was correct here. It should be `file://` to get it to work in the local use case `/`.

actions/src/sema4ai/actions/chat/_client.py
```diff
@@ -156,65 +153,69 @@ def set_bytes(
...
                     (200, 204),
                 )
 
-            # Send the upload complete notification
-            url = f"{self._url}/threads/{thread_id}/files/confirm-upload"
-            headers = {
-                "Content-Type": "application/json",
-                "x-action-context": self.get_x_action_context(),
+            # Confirm upload
+            confirm_url = f"{self._url}/api/v1/threads/{thread_id}/files/confirm-upload"
```
fabioz: The `/api/v1` needs to be passed from the agent server (so, the old thing was correct).

fabioz (approved):
LGTM

PR #135, opened 2024-12-04
First data access template

tonnitommi:
This PR introduces the first Data Access template.

- Uses two data sources: file that is included and a postgres that is hosted and connection details (read only) are shared. This runs in our context aws account. Data sources are linkable - they have shared `customer_id` field.
- Two queries
  - One simple query to one data source only, one paramters
  - Second one is more "real world" that validates the input from the agent, and only if one customer name is found, performs a more complex query that joins two data sources. Also includes proper error handling, and three test inputs. "Like we want people to do".


tonnitommi:
Getting alerts from GitGuardian that a password is exposed here - that is on purpose as otherwise it's pretty hard to make a sample for database connections. :)

kariharju (approved):

PR #150, opened 2025-01-02
Async in UI

fabioz:


OvidiuCode (approved):

action_server/frontend/src/components/CancelButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+type Props = {
+  runId?: string;
+  disabled?: boolean;
+  status: RunStatus;
+  requestId?: string;
+};
+
+export const CancelButton: FC<Props> = ({ runId, disabled, status, requestId }) => {
+  // Either runId or requestId must be provided.
+  if (!runId && !requestId) {
```
el-mike, line 16: since `runId` and `requestId ` are required, I think they shouldn't be optional in the `Props` type?

action_server/frontend/src/components/CancelButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+    throw new Error('Either runId or requestId must be provided.');
```
el-mike, line 17: showing some kind of error state instead of throwing an exception would be less disruptive UX - minor though, as I'm assuming this state should not be reached at all

action_server/frontend/src/lib/requestData.ts
```diff
@@ -333,8 +333,32 @@ export const collectOAuth2Status = async (
...
   });
-  const data = await loadAsync(`${baseUrl}/oauth2/status`, 'GET', {
+  const data = await loadAsync<any>(`${baseUrl}/oauth2/status`, 'GET', {
     params,
   });
   setLoaded(data);
 };
+
+export const cancelRun = async (runId?: string, requestId?: string) => {
+  if (!runId && !requestId) {
```
el-mike, line 343: same - if required, should arguments be made not optional as well? 

If some client deals with potetinal `undefined` values to provide to this function, client could go
```
cancelRun(myPossibleUndefined1 || '', myPossibleUndefined2 || '');
```

action_server/src/sema4ai/action_server/_cli_impl.py
```diff
@@ -905,6 +906,27 @@ def _use_db_ctx(*args, **kwags):
                     if code != 0:
                         return code
 
+                # Ok, we're starting the server. At this point, if there are any
+                # runs in the `not_run` state, or in the `running` state, we should
+                # mark them as cancelled.
+
+                runs_to_cancel = db.all(
```
el-mike, line 913: it feels like having some reusable function for cancelling all pending runs could come in handy ^^ minor ofc

action_server/frontend/src/views/runs/index.tsx
```diff
@@ -32,9 +32,20 @@ import { ActionRunsContext, useActionRunsContext } from './components/context';
...
+  // Issue: if we have a run and click it, we'll set the showRun to the run,
+  // but afterwards, if the run is changed (so, we have a new run instance
+  // representing the same run), we have to use an effect to update the clicked
+  // run.
+
+  useEffect(() => {
+    if (run && showRun && run.id === showRun.id) {
+      setShowRun(run);
+    }
+  }, [run, showRun]);
```
el-mike: would setShowRun ever change? if so, may be worth adding here as well to prevent using stale version of it

action_server/frontend/src/components/CancelButton.tsx
fabioz: Well, one of them is required, but if one is set then the other is optional (thus, both need to be marked as optional in the props).

action_server/frontend/src/lib/requestData.ts
fabioz: It's the same thing, one of them is required while the other is optional.

action_server/frontend/src/components/CancelButton.tsx
fabioz: Actually, this is a programming error -- it's more an assertion error that an invariant in the code is broken than something expected (so, a better UX should really be required).

action_server/src/sema4ai/action_server/_cli_impl.py
fabioz: I guess that if some other place ends up needing this can be refactored (but I'd rather leave as is for now).

action_server/frontend/src/views/runs/index.tsx
fabioz: It shouldn't, but no harm in adding it either, so, I'll change to add that too, who knows if it'll change in the future.

el-mike (approved):

PR #152, opened 2025-01-06
Arm build

fabioz:


action_server/build.py
```diff
@@ -125,36 +120,50 @@ def _resume_download(url: str, filename: str | Path, chunk_size: int = 1024):
...
+        )
+
+    machine = platform.machine()
+    is_64 = not machine or "64" in machine
+
+    if sys.platform == "win32":
+        if is_64:
+            relative_path = "/windows64/rcc.exe"
+        else:
+            relative_path = "/windows32/rcc.exe"
```
kariharju: We don't have 32 bit versions on RCC micromamba etc. so this could be dropped.

action_server/src/sema4ai/action_server/_download_rcc.py
```diff
@@ -47,7 +41,30 @@ def download_rcc(
...
 
-    rcc_url = RCC_URLS[system or platform.system()]
+    machine = platform.machine()
+    is_64 = not machine or "64" in machine
+
+    if sys.platform == "win32":
+        if is_64:
+            relative_path = "/windows64/rcc.exe"
+        else:
+            relative_path = "/windows32/rcc.exe"
```
kariharju: ..same here dropping the 32bit makes this cleaner.

kariharju (approved):
LGTM, we can drop the 32bit parts as we don't have those versions and never will.

action_server/src/sema4ai/action_server/_download_rcc.py
fabioz: Ok, raising exception in this case to make it clear.

action_server/build.py
fabioz: Ok, raising exception in this case to make it clear.

PR #159, opened 2025-01-17
Data access - predict actions template

tonnitommi:
This template introduces three different predictive models including all the datasets as files: classification, linear regression and timeseries forecasting.

While this is much heavier that our typical templates, there is also a lot of complexity with predictive models. The plan with this is that it gives starting point to most common three types of predictions, and user can delete the parts they don't need.

There will be a tutorial in docs that goes through the usage of this template in detail.


templates/data-access-predict/data_actions.py
```diff
@@ -0,0 +1,155 @@
...
+@predict
+def predict_energy_consumption(
+    datasource: EnergyConsumptionModelDataSource,
+    building_type: Optional[str] = None,
+    square_footage: Optional[int] = None,
+    number_of_occupants: Optional[int] = None,
+    appliances_used: Optional[int] = None,
+    average_temperature: Optional[int] = None,
+    day_of_week: Optional[str] = None
+) -> Response[str]:
```
fabioz: Should be `Response[Table]` (same for other predicts)

templates/data-access-predict/data_actions.py
```diff
@@ -0,0 +1,155 @@
...
+    if day_of_week is not None:
+        conditions.append("day_of_week = $day_of_week")
+        params['day_of_week'] = day_of_week
+
+    if conditions:
+        base_query += " WHERE " + " AND ".join(conditions)
+
+    # Finally we execute the query and return the result.
+    result = datasource.query(base_query, params)
+    return Response(result=result.to_markdown())
```
fabioz: Use `result.to_table()` (same for other predicts).

templates/data-access-predict/data_sources.py
```diff
@@ -0,0 +1,73 @@
...
+]
+
+HouseSalesModelDataSource = Annotated[
+    DataSource,
+    DataSourceSpec(
+        name="models",
+        model_name="house_sales_model",
+        engine="prediction:lightwood",
+        description="Please provide a description for this model",
+        setup_sql="CREATE MODEL models.house_sales_model FROM files (SELECT * FROM house_sales) PREDICT MA ORDER BY saledate GROUP BY bedrooms, type -- as the data is quarterly, look back two years to forecast the next one year WINDOW 8 HORIZON 4 -- use the statsforecast engine for this model USING engine = 'statsforecast';",
```
fabioz: I think this SQL should be multi-line formatted (use a string with `"""` --  it's kind of strange to read it from a single line).

tonnitommi: Good point, this is what comes from when we create the predicts with named query creator. I add this also to backlog for the extension to be fixed.

tonnitommi:
Update: I decided to REMOVE one of the predicts and pre-made datasources from the code, leaving only two in the template.

For the third one (Classification) I left in the SQL and the csv file - I am using this example in the documentation as a step by step tutorial how to work with models. It provides much better/realistic experience for the user that the data source and predict are not yet in the code in the template.

templates/data-access-predict/README.md
```diff
@@ -0,0 +1,15 @@
...
+This template shows Sema4.ai Data Access in action with predictive models. It contains all the necessary data files and follows the required structure to get started with building predictive models and creating `@predict` actions that add predictive capabilities to your agent.
+
+Follow the full tutorial using this template in our [documentation](https://sema4.ai/docs/solutions/data-access/models).
+
+## Predictive Models
+
+This example contains three scenarios:
+
+1. **Linear Regression** using the energy consumption data
+2. **Classification** using the customer classification data
```
kariharju: Do we need to drop point 2. as the data source is no there

kariharju (approved):

templates/data-access-predict/README.md
tonnitommi: Changed this to be more clear.

PR #178, opened 2025-03-04
Update advanced template

OvidiuCode:


templates/advanced/src/github/create_actions.py
```diff
@@ -0,0 +1,39 @@
...
+    Creates a new issue in the specified repository.
+
+    Args:
+        github_access_token: Your private GitHub access token.
+        repository_info: Information needed to locate the repository.
+        issue_data: Information about the issue to create.
+        limit: The maximum number of commits to return (max 100).
+
+    Returns:
+        A list of commit messages.
```
fabioz: Return message seems misleading since it returning a `Response[str]`.

templates/advanced/tests/test_template.py
```diff
@@ -0,0 +1,5 @@
+def test_actions() -> None:
+    from github.get_actions import get_repository_commits  # noqa
```
fabioz, line 2: Why the `noqa`? It seems like these imports should work...

templates/advanced/package.yaml
```diff
@@ -12,9 +12,25 @@ dependencies:
     - python=3.10.14
     - uv=0.4.17
   pypi:
-    - sema4ai-actions=1.3.0
-    - pydantic=2.10.4
+    - sema4ai-actions=1.3.5
+    - pydantic=2.10.6
     - requests=2.32.3
+    - ruff=0.9.9
```
fabioz: ruff and mypy should be dev-dependencies

templates/advanced/tests/test_template.py
OvidiuCode: Needed for lint to pass (import not used)

fabioz (approved):

PR #183, opened 2025-03-11
Filename is required in multipart file upload

mstn:
## Context

It seems that the [library](https://www.npmjs.com/package/multer) we use for uploading files does not "extract" the file from the multipart body request if `filename` is not specified.. 

A [quick look](https://github.com/expressjs/multer/blob/master/lib/make-middleware.js#L99) reveals that if the filename is not specified in the multipart request, the file uploading is possibly skipped entirely. 

Even if we should be probably more flexible on the backend side, for the sake of time, we prefer to patch the file actions and provide the required field.

[Discussion on Slack](https://sema4ai.slack.com/archives/C084H2PSZTK/p1741701060702769)

## More Evidence

We tested the behaviour with an agent.

TLTR
* ` fields.append(("file", ("hello.txt", "Hello!")))` OK
*  `fields.append(("file", "Hello!"))` NOT OK.

[agent-package.zip](https://github.com/user-attachments/files/19195206/agent-package.zip)

### DIFF
--- a/actions/src/sema4ai/actions/chat/_client.py
+++ b/actions/src/sema4ai/actions/chat/_client.py
@@ -211,7 +211,8 @@ files will be stored."""
                 # Add the file to the request after the form data (the order
                 # of the fields is important -- the "key" field must
                 # be the first field).
-                fields.append(("file", content))
+                filename = "file"
+                fields.append(("file", (filename, content)))
 
                 encoded_data, content_type = encode_multipart_formdata(fields)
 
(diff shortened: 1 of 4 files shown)


fabioz (approved):

actions/src/sema4ai/actions/chat/_client.py
```diff
@@ -211,7 +211,8 @@ def set_bytes(
                 # Add the file to the request after the form data (the order
                 # of the fields is important -- the "key" field must
                 # be the first field).
-                fields.append(("file", content))
+                filename = "file"
```
fabioz: Actually, the `filename` is in the namespace, so, maybe it should be `use_filename = filename`.

fabioz: Or just remove the assign line ;)

PR #184, opened 2025-03-12
`sema4ai.common.app_mutex.obtain_app_mutex` api

fabioz:


OvidiuCode (approved):

common/src/sema4ai/common/process.py
```diff
@@ -440,6 +440,51 @@ def kill_process_and_subprocesses(pid):
...
+
+        log.info(f"Killing processes: {names}")
+        for p in children_processes:
+            try:
+                p.kill()
+            except Exception as e:
+                log.debug(f"Exception when terminating process: {p.pid}: {e}")
+
+        # Give processes 2 seconds to exit cleanly and force-kill afterwards
+        _gone, alive = psutil.wait_procs(children_processes, 2)
```
kylie-bee: @fabioz can we make this grace period configurable? The binary does quite a bit of shut down work and on slower machines it might take longer than 2 seconds

PR #193, opened 2025-03-26
Use sema4ai http helper for requests

OvidiuCode:


action_server/tests/action_server_tests/test_server_https.py
```diff
@@ -19,12 +20,15 @@ def test_serve_https(action_server_process: ActionServerProcess):
     )
 
     client = ActionServerClient(action_server_process, use_https=True)
-    openapi_json = client.get_openapi_json()
-    json.loads(openapi_json)  # Just check it can be loaded
+    requests.get(
```
fabioz, line 23: Why is this change done? The old code should do almost the same, when you pass `use_https=True`, no?

fabioz: Just saw that this support was removed. I think you shouldn't remove this support, rather, use the proper idioms in the sema4ai http helper...

action_server/src/sema4ai/action_server/_selftest.py
```diff
@@ -265,30 +265,27 @@ def requests_kwargs(self) -> dict:
         """
         Provides the default kwargs that should be used for a requests call.
         """
-        from sema4ai.action_server._settings import get_user_sema4_path
-
         kwargs: dict = dict(timeout=self._get_default_timeout())
-        if self._use_https:
```
fabioz, line 271: Doesn't sema4ai http helper have a way to verify a certificate?

fabioz (approved):

PR #195, opened 2025-03-26
Fix DBError in execute_update_returning method

joshyorko:
Address `DBError` in `execute_update_returning` method by managing transactions correctly.

* **Transaction Management in `execute_update_returning`**:
  - Add a transaction initiation using `BEGIN` before executing the `UPDATE` statement.
  - Add a `COMMIT` statement after executing the SQL statements.
  - Add a `ROLLBACK` statement to handle errors.

* **Transaction Management in `_create_run`**:
  - Add a check using `in_transaction()` to ensure a transaction is active before calling `execute_update_returning`.
  - Add a transaction initiation using `BEGIN` before calling `execute_update_returning`.
  - Add a `COMMIT` statement after calling `execute_update_returning`.
  - Add a `ROLLBACK` statement to handle errors.

### DIFF
--- a/action_server/src/sema4ai/action_server/_actions_run.py
+++ b/action_server/src/sema4ai/action_server/_actions_run.py
@@ -72,21 +72,28 @@ def _create_run(
         request_id=request_id,
     )
     with db.transaction():
-        with db.cursor() as cursor:
-            db.execute_update_returning(
-                cursor,
-                "UPDATE counter SET value=value+1 WHERE id=? RETURNING value",
-                [RUN_ID_COUNTER],
-            )
-            counter_record = cursor.fetchall()
-            if not counter_record:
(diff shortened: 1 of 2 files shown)


fabioz:
Sorry for the long delay in getting back to you (it seems that I wasn't getting notifications properly for some reason, so, going over existing PRs now).

I'm not sure I understand the issue (since the `db.transaction()` context manager should be handling this -- https://github.com/Sema4AI/actions/blob/master/action_server/src/sema4ai/action_server/_database.py#L241).

Can you create an actual test case in https://github.com/Sema4AI/actions/blob/master/action_server/tests/action_server_tests/test_database.py that shows the error to be fixed? (my impression is that the fix, if any, must be in the context manager / database module, not here).

PR #197, opened 2025-03-27
Use uv to setup python environment (to get around issue where the pyt…

fabioz:
…hon version was built with an old version of sqlite in ubuntu 20.04).


.github/workflows/_gen_workflows.py
kariharju, line 1: Need really good guidance and a link in readme to that guidance on where and how to edit the GHA scripts.
Be default people like me will go into the GHA files and edit them.
- Add big comment block in the generated files: "Auto-generated files guidance at: <gha.md>"

kariharju (approved):

OvidiuCode (approved):

.github/workflows/_gen_workflows.py
fabioz: Done

PR #200, opened 2025-03-31
Add support for network-settings.yaml file

OvidiuCode:
network-settings.yaml example:

```yaml
spec-version: 1.0.0

current-profile: Home
profiles:
  - name: "Home"
    description: "Default home network settings"
    disable-ssl: false                   # By default traffic should always be HTTPS
    
  - name: "Work"
    description: "My proxy settings for work"
    disable-ssl: false
    proxy-settings:
      authentication: 
        auth-type: os-certificate
      no-proxy: "localhost,127.0.0.1"
      https-proxy: https://work-proxy.company.com
      http-proxy: http://work-proxy.company.com
  
  - name: "Insecure Proxy"  # Example of a profile with all SSL disable, proxy is there with no auth
    description: "Direct connection without proxy"
    disable-ssl: true
    proxy-settings:
      authentication: 
        auth-type: none
      no-proxy: ""
      https-proxy: http://no-ssl-proxy.company.com
      http-proxy: http://no-ssl-proxy.company.com
  ```

### CODE REVIEW

File: sema4ai-http-helper/pyproject.toml [hunk 1]
```diff
@@ -14,6 +14,8 @@ packages = [{ include = "*", from = "src" }]
 python = "^3.10"
 truststore = "0.9.2"
 urllib3 = ">=2.2.2"
+pyyaml = "6.0.2"
```
fabioz: Given that this is a helper to be included in other projects, make this more flexible (`>=5.0` would work).

sema4ai-http-helper/src/sema4ai_http/__init__.py
```diff
@@ -14,60 +16,169 @@
...
-    ALREADY_EXISTS = auto()
-    HTTP_ERROR = auto()
+    def _build_ssl_context(
+        self,
+        protocol: int | None,
+        enable_legacy_server_connect: bool | None = None,
+    ) -> ssl.SSLContext:
+        import truststore
 
+        ssl_context = truststore.SSLContext(protocol or ssl.PROTOCOL_TLS_CLIENT)
```
fabioz: 0 could be a valid protocol (since it's an int), so, use the protocol checking for None instead of `or`.

sema4ai-http-helper/src/sema4ai_http/__init__.py
```diff
@@ -14,60 +16,169 @@
...
+
+        return sema4_home / "network-settings.yaml"
+
+    def _get_current_config_profile(self) -> dict:
+        config_file = self._get_network_settings_path()
+
+        if config_file.exists():
+            import yaml
+
+            config = yaml.safe_load(config_file.read_text())
```
fabioz: Give a proper error message when this fails (should have 1 error for failure in `read_text` and another for `yaml.safe_load`).

sema4ai-http-helper/src/sema4ai_http/__init__.py
```diff
@@ -14,60 +16,169 @@
...
+            current_profile = config.get("current-profile", "")
+            return next(
+                (p for p in config.get("profiles", []) if p["name"] == current_profile),
```
fabioz: If the user does a bad thing and `p['name']` isn't there, this will raise a keyerror (so, either log and ignore or raise a better error message).

sema4ai-http-helper/src/sema4ai_http/__init__.py
fabioz: Also, verify that `config` is actually a dict...

sema4ai-http-helper/src/sema4ai_http/__init__.py
```diff
@@ -14,60 +16,169 @@
...
-        cert_reqs="CERT_REQUIRED",
+def build_ssl_context(
+    protocol: int | None = None, *, enable_legacy_server_connect: bool | None = None
+) -> "ssl.SSLContext":
+    return SSLContextFactory({})._build_ssl_context(
+        protocol, enable_legacy_server_connect
     )
 
 
+class ResponseWrapper:
```
fabioz, line 151: This should be a public facing class, as such, please add proper docs to it.

A TODO I had is actually document existing methods (3rd party clients have no way of knowing how the library works right now)... if you have the time, it'd be nice to fix that too (if not, that's ok, but let's try to document the API we're actually making public now).

sema4ai-http-helper/src/sema4ai_http/__init__.py
```diff
@@ -14,60 +16,169 @@
...
+class DownloadStatus(Enum):
+    """Status"""
+
+    DONE = auto()
+    PARTIAL = auto()
+    ALREADY_EXISTS = auto()
+    HTTP_ERROR = auto()
+
+
+connection_manager = NetworkConfig().connection_pool
```
fabioz: Make this lazy (i.e.: make it a method with `lru_cache`. Also, do we want to make that public? I'd just make it private for now -- i.e.: `_get_connection_manager`).

sema4ai-http-helper/src/sema4ai_http/__init__.py
OvidiuCode: great point

.github/workflows/common_tests.yml
```diff
@@ -11,26 +11,25 @@
...
+"on":
   push:
     branches:
-    - master
-    - wip
+      - master
+      - wip
     paths:
-    - common/**
-    - .github/workflows/common_tests.yml
```
fabioz, line 23: You shouldn't need to change this file... if you revert these changes,  do you still get errors?

fabioz (approved):

PR #208, opened 2025-04-18
Fix: Queries on tables with array columns fail

rahul-dange-talentica:
added support for list and dict datatypes for table columns

### DIFF
--- a/actions/docs/CHANGELOG.md
+++ b/actions/docs/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.3.9 - 2025-04-18
+
+- Added support for list and dict datatypes for table columns
+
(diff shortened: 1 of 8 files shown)


fabioz:
It seems tests are failing due to missing doc updates.

Running `inv docs` and pushing the result should fix this.

fabioz:
Also, please add an entry to the CHANGELOG.md with details on the change.

fabioz (approved):

fabioz:
It seems this came up again and was fixed in another commit (https://github.com/Sema4AI/actions/commit/450dfe9636c1eb4a92bef7306cac5d284560ab54) as no one remembered about this PR (which was approved but not merged).

So, closing this PR now.

PR #210, opened 2025-04-24
Update to the latest robocorp-log version

OvidiuCode:


actions/pyproject.toml
```diff
@@ -11,7 +11,7 @@ packages = [{ include = "sema4ai/actions", from = "src" }]
 [tool.poetry.dependencies]
 python = "^3.10"
 cryptography = "^43.0.1"
-robocorp-log = ">=2.4,<3"
```
fabioz, line 14: Given that the API is still the same, we could do:

`>=2.4,<4`

fabioz (approved):

PR #211, opened 2025-04-24
Add `get_network_profile` method

vlad-sema4:


sema4ai-http-helper/src/sema4ai_http/__init__.py
```diff
@@ -143,11 +146,51 @@ def _get_connection_manager() -> urllib3.PoolManager | urllib3.ProxyManager:
...
+
+    @classmethod
+    def from_network_config(cls, network_config: _NetworkConfig) -> "NetworkProfile":
+        return cls(
+            ssl_context=network_config.get_ssl_context(),
+            proxy_config=ProxyConfig.from_network_config(network_config),
+        )
+
+
+def get_network_profile() -> NetworkProfile | None:
```
OvidiuCode: the return type will never be None here, right?

sema4ai-http-helper/src/sema4ai_http/__init__.py
```diff
@@ -58,20 +58,8 @@ class _NetworkConfig:
...
-            sema4_home = Path(localappdata) / "sema4ai"
-        else:
-            # Linux/Mac
-            sema4_home = Path("~/.sema4ai").expanduser()
-
-        return sema4_home / "network-settings.yaml"
-
-    def _get_current_config_profile(self) -> dict:
+    @cached_property
+    def profile_config(self) -> dict:
```
fabioz: Do we know the expected shape of the dict here (maybe a TypedDict would be better if we know about it).

OvidiuCode (approved):

OvidiuCode (approved):

PR #213, opened 2025-04-25
Add guide and missing get_file docs for chat module

OvidiuCode:


actions/docs/guides/02-chat-files.md
```diff
@@ -0,0 +1,83 @@
+## Working with Files in Chat Actions
+
+The `sema4ai.actions.chat` module provides functionality to attach and retrieve files during chat interactions.
+
+### Default File Storage Location
+
+By default, files are stored in a local directory at:
```
fabioz: I think it should explain a bit better that in dev we use the local directory in production workroom will store those in the cloud.

Also, it should be clearer that only in VSCode with our extension the file is saved at the devdata (i.e.: it's not just by default, it's by default in VSCode when running with the `Sema4.ai Extension`).

actions/docs/guides/02-chat-files.md
```diff
@@ -0,0 +1,83 @@
+## Working with Files in Chat Actions
```
fabioz: We should also add a section saying why this is needed (something along the lines that when running in the cloud, accessing anything in the local file system isn't possible and these APIs should be used to access anything that needs to be thread-scoped).

fabioz (approved):

PR #219, opened 2025-05-30
Add dev-tasks to all templates

OvidiuCode:


templates/minimal/package.yaml
```diff
@@ -14,6 +14,22 @@ dependencies:
...
 
+dev-dependencies:
+  conda-forge:
+    - pytest=8.3.5
+    - types-requests=2.32.0.20250301
+    - ruff=0.11.12
+    - mypy=1.16.0
+
+dev-tasks:
+  test: pytest tests
```
kariharju, line 25: For the minimal I don't think we can bring pytests..
Linters and ruff checks yes but pushing the whole shebang right from the most minimal setup feels a bit lot.

fabioz: Humm, I don't know -- for me the minimal thing to have is a project with tests/lint/type-checking, so, I'd keep it...

templates/minimal/tests/test_template.py
```diff
@@ -0,0 +1,4 @@
+def test_actions() -> None:
+    from actions import greet  # noqa
+
+    assert 1 == 1, "TODO: Add your tests here"
```
fabioz: I think I'd call greet here and check its result (and remove the `noqa`).

fabioz (approved):

templates/minimal/tests/test_template.py
fabioz: i.e.: this would show users they can just call any `@action` directly for testing without any issues...

PR #223, opened 2025-06-04
Remove predict template

kariharju:
The template has not been visible for awhile now but the repo is public and has the template and docs on it so dropping will clean things up.


kariharju:
Removed elsewhere

PR #224, opened 2025-06-05
Add support for using the MCP protocol to list/call actions as tools in the MCP protocol.

fabioz:


action_server/src/sema4ai/action_server/_actions_run.py
```diff
@@ -171,6 +171,36 @@ def _set_run_as_running(run: "Run", initial_time: float) -> int:
...
+
+class ResponseHandler:
+    def __init__(self, response: Response):
+        self.response = response
+
+    def set_run_id(self, run_id: str):
+        self.response.headers["X-Action-Server-Run-Id"] = run_id
+
+    def set_async_completion(self):
+        self.response.headers["x-action-async-completion"] = "1"
```
cosmogp, line 190: Can we keep these consistent - lowercased or capitalized?
As they are case-insensitive by default, this would mean only code consistency.
Or are they following some convention? If so, a comment would help.

cosmogp: Also, the magic strings might be better as constants.

action_server/src/sema4ai/action_server/_actions_run.py
```diff
@@ -611,7 +649,7 @@ async def func(response: Response, request: Request):
         # This is done because headers have a size restriction and we don't want to
         # hit it (so, we enable passing things that are conceptually headers, such as
         # `x-action-context` and `x-data-context`, in the body of the request)
-        invocation_context = request.headers.get("x-action-invocation-context")
+        invocation_context = headers.get("x-action-invocation-context")
```
cosmogp, line 652: Magic strings might be better as constants.

action_server/src/sema4ai/action_server/_actions_run.py
```diff
@@ -596,10 +630,14 @@ async def func(response: Response, request: Request):
...
             )
-
         headers = dict((x[0].lower(), x[1]) for x in request.headers.items())
         cookies = dict(request.cookies)
+        response_handler = ResponseHandler(response)
+        return await func_internal(response_handler, inputs, headers, cookies)
 
+    async def func_internal(
+        response_handler: IResponseHandler, inputs, headers, cookies
+    ):
```
cosmogp, line 640: A bit harsh on readability... inputs, headers, cookies don't have a straightforward indication of where they are coming from

action_server/src/sema4ai/action_server/_api_action_routes.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+def _name_to_url(name: str) -> str:
+    from sema4ai.action_server._slugify import slugify
+
+    return slugify(name.replace("_", "-"))
+
+
+def build_url_api_run(action_package_name: str, action_name: str) -> str:
+    return f"/api/actions/{_name_to_url(action_package_name)}/{_name_to_url(action_name)}/run"
```
cosmogp, line 25: Maybe something like might be a better option `urllib.parse.urljoin`

action_server/src/sema4ai/action_server/_api_action_routes.py
```diff
@@ -0,0 +1,340 @@
...
+                if key.lower() == "authorization"
+            ),
+            None,
+        )
+        if not auth_header or not auth_header.lower().startswith("bearer "):
+            raise HTTPException(
+                status_code=HTTP_403_FORBIDDEN, detail="Not authenticated"
+            )
+
+        token = auth_header[7:]  # Remove "Bearer " prefix
```
cosmogp, line 75: It might be better to split based on spaces then just removing the characters.
This way you avoid multiple spaces even though header starts with the specific string.

action_server/src/sema4ai/action_server/_api_action_routes.py
```diff
@@ -0,0 +1,340 @@
...
+        if api_key:
+            middleware.append(
+                Middleware(
+                    AuthenticationMiddleware,
+                    backend=APIKeyAuthBackend(api_key=api_key),
+                )
+            )
+
+        self._setup_mcp(app, middleware)
+        self._setup_sse(app, middleware)
```
cosmogp, line 136: Do you need to setup the SSE in the `setup_mcp_server` function?
Maybe rename function?

action_server/src/sema4ai/action_server/_api_action_routes.py
```diff
@@ -0,0 +1,340 @@
...
+            (action_package.id, action_package)
+            for action_package in db.all(ActionPackage)
+        )
+
+        actions = db.all(Action)
+        registered_route_names: set[str] = set()
+        for action in actions:
+            if not action.enabled:
+                # Disabled actions should not be registered.
+                continue
```
cosmogp, line 251: I don't remember, is this a feature in the decorator?

action_server/src/sema4ai/action_server/_api_action_routes.py
```diff
@@ -0,0 +1,340 @@
...
+                func_fast_api,
+                func_internal,
+                openapi_extra,
+            ) = _actions_run.generate_func_from_action(
+                action_package, action, display_name
+            )
+
+            if action.is_consequential is not None:
+                openapi_extra["x-openai-isConsequential"] = action.is_consequential
+            openapi_extra["x-operation-kind"] = action_kind
```
cosmogp, line 301: Maybe magic strings could be constants

action_server/src/sema4ai/action_server/_api_action_routes.py
```diff
@@ -0,0 +1,340 @@
...
+                from ._whitelist import accept_action
+
+                if not accept_action(self.whitelist, action_package.name, action.name):
+                    log.info(
+                        "Skipping action %s / %s (not in whitelist)",
+                        action_package.name,
+                        action.name,
+                    )
+                    continue
+            display_name = _name_as_summary(action.name)
```
cosmogp, line 279: I'm a bit confused by this.
Is the name converted to a summary? How can a name be the summary?

action_server/src/sema4ai/action_server/_api_action_routes.py
cosmogp, line 136: It looks like we're still doing the setup for MCP + SSE if the API key is None.
If the authentication is mandatory for those, maybe throw an exception?

el-mike (approved):
Amazing job @fabioz ! Looks great, nothing caught my eye (although I'm that that familiar with Action Server implementation ^^)

action_server/src/sema4ai/action_server/_actions_run.py
fabioz: Done

action_server/src/sema4ai/action_server/_actions_run.py
fabioz: Done

action_server/src/sema4ai/action_server/_actions_run.py
fabioz: I added some docstrings, hopefully it's more understandable.

action_server/src/sema4ai/action_server/_api_action_routes.py
fabioz: I'd like to leave this as is.

action_server/src/sema4ai/action_server/_api_action_routes.py
fabioz: I think that changing to that would actually be a bug. 

I may be wrong, but I believe that it must be: `Bearer` (case insensitive) then 1 space and the token, if the client does it wrong it should fail (checked chatgpt and it seems to agree, but I didn't go read the actual oauth RFCs to double-check).

action_server/src/sema4ai/action_server/_api_action_routes.py
fabioz: I'll rename these functions to `_setup_mcp_streamable_route` and `_setup_mcp_sse_route` to make it clearer -- the `/mcp` is the `streamable` route and the `/sse` is the server side events route, so, when setting up `mcp server` we're making both routes available (a websocket may follow in the future).

action_server/src/sema4ai/action_server/_api_action_routes.py
fabioz: No... if you import an action it'll always be there, but it's possible to resync in the future and upon resynchronizing the actions, if an action was removed it'll be marked as "disabled" (or someone could call `action-server datadir clear-actions` to clear all actions which marks all actions as disabled internally and then reimport passing a `whitelist`).

action_server/src/sema4ai/action_server/_api_action_routes.py
fabioz: Done

action_server/src/sema4ai/action_server/_api_action_routes.py
fabioz: I renamed that function to `_make_name_user_friendly`, the function name was a bit misleading.

action_server/src/sema4ai/action_server/_api_action_routes.py
fabioz: It's not mandatory, if the API key is not set, we should still setup the mcp server.

PR #236, opened 2025-06-19
When loading actions, all python files are now checked for the decorators in the contents instead of just loading specific files such as `action.py`

fabioz:


OvidiuCode (approved):

actions/src/sema4ai/actions/_collect_actions.py
```diff
@@ -12,6 +13,149 @@
...
+        ".venv",
+        "venv",
+        "env",
+        "build",
+        "dist",
+        ".pytest_cache",
+        ".tox",
+        "site-packages",
+        "pip-global",
+    }
```
cosmogp: Do we have a way of giving the user the ability of filtering?
Like a gitignore for files that should be ignored when gathering available actions?

actions/src/sema4ai/actions/_collect_actions.py
```diff
@@ -12,6 +13,149 @@
...
+    def _matches_glob_patterns(self, path: Path) -> bool:
+        """Check if a path matches any of the glob patterns."""
+        for pattern in self.glob_patterns:
+            # Simple glob pattern matching (just based on the name, not full path -- it's an 'rglob' by default)
+            if fnmatch.fnmatch(path.name, pattern):
+                return True
+        return False
+
+    def _iterate_paths_recursive(
+        self, current_path: Path, depth: int = 0
```
cosmogp: Maybe magic numbers into constants?

actions/src/sema4ai/actions/_collect_actions.py
```diff
@@ -12,6 +13,149 @@
...
+
+            try:
+                for item in current_path.iterdir():
+                    if item.is_file():
+                        # Check files at current depth
+                        if item.suffix == ".py" and self._matches_glob_patterns(item):
+                            yield item
+                    elif item.is_dir():
+                        # Only recurse into directories if we haven't reached the depth limit
+                        if depth < 3:
```
cosmogp: Maybe magic numbers into constants?

actions/src/sema4ai/actions/_collect_actions.py
```diff
@@ -12,6 +13,149 @@
...
+
+            # Skip common directories
+            if current_path.name in self.SKIP_DIRS:
+                return
+
+            try:
+                for item in current_path.iterdir():
+                    if item.is_file():
+                        # Check files at current depth
+                        if item.suffix == ".py" and self._matches_glob_patterns(item):
```
cosmogp: There's no other python file suffix that we need to cover?
Just asking...

actions/src/sema4ai/actions/_collect_actions.py
```diff
@@ -12,6 +13,149 @@
...
+    if hasattr(site, "getusersitepackages"):
+        site_paths = site.getusersitepackages()
+        if isinstance(site_paths, (list, tuple)):
+            for site_path in site_paths:
+                roots.append(site_path)
+        else:
+            roots.append(site_paths)
+
+    if hasattr(site, "getsitepackages"):
+        site_paths2 = site.getsitepackages()
```
cosmogp: Maybe better name for these?

cosmogp (approved):

actions/src/sema4ai/actions/_collect_actions.py
fabioz: No, we just accept actions in .py files.

actions/src/sema4ai/actions/_collect_actions.py
fabioz: I think we could use the `package.yaml` exclude rules here.

actions/src/sema4ai/actions/_collect_actions.py
fabioz: I think in this case (depth = 0) a constant for the initial value may be a bit too much.

actions/src/sema4ai/actions/_collect_actions.py
fabioz: I changed to use the `package.yaml` exclusion rules.

actions/src/sema4ai/actions/_collect_actions.py
fabioz: Created constant

actions/src/sema4ai/actions/_collect_actions.py
fabioz: Done

PR #240, opened 2025-06-23
DEV-1949: Resolve dependency conflict on psutil for browser-use package

cosmogp:
[DEV-1949: Resolve dependency conflict on psutil for browser-use package](https://linear.app/sema4ai/issue/DEV-1949/resolve-dependency-conflict-on-psutil-for-browser-use-package)

### DIFF
--- a/actions/docs/CHANGELOG.md
+++ b/actions/docs/CHANGELOG.md
@@ -1,7 +1,6 @@
 # Changelog
 
 ## 2.0.0 - 2025-06-23
-
 - The heuristics for finding actions/tools/etc. have been changed (**Backward Incompatible change**).
   - The new heuristics are:
     - Any python file will now be considered for having an action defined in it.
--- a/actions/docs/CHANGELOG.md
+++ b/actions/docs/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 2.0.0 - 2025-06-23
 
 - The heuristics for finding actions/tools/etc. have been changed (**Backward Incompatible change**).
   - The new heuristics are:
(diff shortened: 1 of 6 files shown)


actions/docs/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
```
fabioz, line 3: This version is backward-incompatible, so, it must be a 2.0.

Note: please don't release yet, I also want to merge https://github.com/Sema4AI/actions/pull/238 which may also be backward incompatible...

actions/pyproject.toml
```diff
@@ -13,16 +13,16 @@ python = "^3.10"
 cryptography = "^43.0.1"
```
fabioz: It seems `cryptography` also has an update, so, it could be:

`>=43.0.1,<46.0.0`

actions/pyproject.toml
```diff
@@ -10,10 +10,10 @@ packages = [{ include = "sema4ai/actions", from = "src" }]
 
 [tool.poetry.dependencies]
 python = "^3.10"
-cryptography = "^43.0.1"
+cryptography = ">=43.0.1,<46.0.0"
 robocorp-log = ">=2.4,<4"
```
fabioz, line 14: Changes here look good, but I think we need to release `robocorp-log` first and then update here so that the latest `psutil` is in the `poetry.lock` for `sema4ai-actions`.

cosmogp: @fabioz I updated this PR: https://github.com/robocorp/robocorp/pull/446
Does this look OK? Let me know.

PR #241, opened 2025-06-24
Add helper functions to interact with the agent server public api

OvidiuCode:


actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+    CONNECTION_TIMEOUT = 1
+
+    def __init__(self, api_key: str | None = None):
+        """Initialize the AgentServerClient."""
+        self.api_url = self._get_api_url()
+        self.api_key = api_key
+        log.info(f"API URL: {self.api_url}")
+
+        # Determine if we're running in cloud environment
+        self.is_cloud = self._is_cloud_environment()
```
fabioz: I don't think you need to have this distiction... rather, if the api_key is passed, always pass the header, otherwise don't.

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+
+import sema4ai_http
+
+log = logging.getLogger(__name__)
+
+
+class _AgentAPIClient:
+    API_ENDPOINT_VERSIONS = ["v1", "v2"]
+    API_PATH_PREFIX = "api/public"
+    ENV_VAR_NAME = "SEMA4AI_AGENTS_SERVICE_URL"
```
fabioz: `ENV_VAR_NAME` doesn't sem a good name for a variable... maybe it could be more descriptive (say `SEMA4AI_AGENT_SERVICE_URL_ENV_VAR_NAME`).

As a note, if it's used just in one place, maybe I wouldn't even create a separate constant...

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+from urllib.parse import urljoin, urlparse
+
+import sema4ai_http
+
+log = logging.getLogger(__name__)
+
+
+class _AgentAPIClient:
+    API_ENDPOINT_VERSIONS = ["v1", "v2"]
+    API_PATH_PREFIX = "api/public"
```
fabioz: Maybe `PUBLIC_API_URL_PREFIX` would be a better name?

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+                return endpoint_url
+        return None
+
+    def _is_url_accessible(self, url: str) -> bool:
+        try:
+            parsed_url = urlparse(url)
+            if parsed_url.scheme not in ("http", "https"):
+                return False
+
+            sema4ai_http.get(url, timeout=self.CONNECTION_TIMEOUT)
```
fabioz: You're not really checking the result here, so, I don't think this will fail as it should... also, I think you want to pass the api key to get the real result.

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Returns:
+            str: Path to the PID file
+        """
+        home_dir = os.path.expanduser("~")
+
+        # Determine OS-specific path
+        if platform.system() == "Windows":
+            # Windows path: C:\Users\<username>\AppData\Local\sema4ai\sema4ai-studio\agent-server.pid
+            return os.path.join(
```
fabioz, line 124: This is not the proper way to get the LOCALAPPDATA. Use the related env var instead.

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+        elif method == "DELETE":
+            response = sema4ai_http.delete(url, headers=request_headers)
+        else:
+            raise ValueError(f"Unsupported HTTP method: {method}")
+
+        response.raise_for_status()
+
+        return response
+
+    def _is_cloud_environment(self) -> bool:
```
fabioz: I think you should remove this function as the functionality it's used for shouldn't be needed.

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+
+from sema4ai.actions.agent._client import _AgentAPIClient
+
+log = logging.getLogger(__name__)
+
+
+class AgentApiClientException(Exception):
+    """Exception raised when the Agent API client encounters an error."""
+
+    pass
```
fabioz: Minor: `pass` is not needed if you have a docstring...

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+        if not paginated_response.next or not paginated_response.has_more:
+            break
+
+        next_token = paginated_response.next
+
+    return all_data
+
+
+def _get_conversations(client: _AgentAPIClient, endpoint: str) -> list[Conversation]:
+    full_url = f"{client.api_url}/{endpoint}"
```
fabioz: Using `urljoin` may be safer...

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+    client = _AgentAPIClient(api_key=sema4_api_key)
+
+    endpoint = f"agents/{agent_id}/conversations"
+    log.info(f"Agent Server API Call URL: {endpoint}")
+
+    return _get_conversations(client, endpoint)
+
+
+def get_conversation(
+    agent_name: str, conversation_name: str, sema4_api_key: str | None = None
```
fabioz: How does one "know" about the `sema4_api_key` he should use?

fabioz (commented):
This is just a partial review (sorry, need more time to review the rest).

actions/docs/api/sema4ai.actions.agent.md
```diff
@@ -0,0 +1,217 @@
...
+
+## `create_conversation`
+
+Creates a new conversation for communication with an agent.
+
+**Args:**
+
+- <b>`agent_id`</b>: The id of the agent to create conversation with
+- <b>`conversation_name`</b>: The name of the conversation to be created
+- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use LOCAL if in Studio or SDK!
```
cosmogp:
```suggestion
- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use "LOCAL" as value if in Studio or SDK!
```

actions/docs/api/sema4ai.actions.agent.md
```diff
@@ -0,0 +1,217 @@
...
+
+**Returns:**
+The created conversation.
```
cosmogp:
```suggestion
The created conversation object.
```

actions/docs/api/sema4ai.actions.agent.md
```diff
@@ -0,0 +1,217 @@
...
+
+______________________________________________________________________
+
+## `get_all_agents`
+
+Fetches a list of all available agents with their IDs and names.
+
+**Args:**
+
+- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Leave empty if in Studio or SDK!
```
cosmogp:
```suggestion
- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use None if in Studio or SDK!
```

actions/docs/api/sema4ai.actions.agent.md
```diff
@@ -0,0 +1,217 @@
...
+______________________________________________________________________
+
+## `get_agent_by_name`
+
+Fetches the agent that matches the name.
+
+**Args:**
+
+- <b>`name`</b>: The name of the agent
+- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Leave empty if in Studio or SDK!
```
cosmogp:
```suggestion
- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use None if in Studio or SDK!
```

actions/docs/api/sema4ai.actions.agent.md
```diff
@@ -0,0 +1,217 @@
...
+______________________________________________________________________
+
+## `get_conversations`
+
+Fetches all conversations for an agent.
+
+**Args:**
+
+- <b>`agent_id`</b>: The ID of the agent
+- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Leave empty if in Studio or SDK!
```
cosmogp:
```suggestion
- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use None if in Studio or SDK!
```

actions/docs/api/sema4ai.actions.agent.md
```diff
@@ -0,0 +1,217 @@
...
+
+## `get_conversation`
+
+Fetches the conversation with the given name for an agent.
+
+**Args:**
+
+- <b>`agent_name`</b>: The name of the agent
+- <b>`conversation_name`</b>: The name of the conversation
+- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Leave empty if in Studio or SDK!
```
cosmogp:
```suggestion
- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use None if in Studio or SDK!
```

actions/docs/api/sema4ai.actions.agent.md
```diff
@@ -0,0 +1,217 @@
...
+
+## `get_conversation_messages`
+
+Fetches all messages from a specific conversation.
+
+**Args:**
+
+- <b>`agent_id`</b>: The ID of the agent
+- <b>`conversation_id`</b>: The ID of the conversation
+- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use LOCAL if in Studio or SDK!
```
cosmogp:
```suggestion
- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use "LOCAL" as value if in Studio or SDK!
```

actions/docs/api/sema4ai.actions.agent.md
```diff
@@ -0,0 +1,217 @@
...
+## `send_message`
+
+Sends a message within a conversation and retrieves the agent's response.
+
+**Args:**
+
+- <b>`conversation_id`</b>: The ID of the conversation
+- <b>`agent_id`</b>: The ID of the agent to send message to
+- <b>`message`</b>: The message content to send
+- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use LOCAL if in Studio or SDK!
```
cosmogp:
```suggestion
- <b>`sema4_api_key`</b>: The API key for the Sema4 API if running in cloud. Use "LOCAL" as value if in Studio or SDK!
```

actions/docs/api/sema4ai.actions.agent.md
cosmogp: Looking at all of these, couldn't we just use the fact that if sema4_api_key is None than is Studio or SDK & if it has a non-empty string value then it's Cloud?

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+            if parsed_url.scheme not in ("http", "https"):
+                return False
+
+            headers = (
+                {"Authorization": f"Bearer {self.api_key}"} if self.api_key else None
+            )
+            sema4ai_http.get(url, headers=headers, timeout=1).raise_for_status()
+            return True
+        except Exception:
+            return False
```
cosmogp, line 106: Maybe log the exception?

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+            return os.path.join(
+                local_app_data,
+                "sema4ai",
+                "sema4ai-studio",
+                self.PID_FILE_NAME,
+            )
+        else:
+            # macOS/Linux path: ~/.sema4ai/sema4ai-studio/agent-server.pid
+            return os.path.join(
+                os.path.expanduser("~"),
```
cosmogp, line 133: Just an observation... This seems a bit strange, I've always steered away from these expressions as in my head they work as language that needs to be decoded

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Returns:
+            sema4ai_http.ResponseWrapper object
+
+        Raises:
+           ValueError: for unsupported HTTP methods
+           AgentApiClientException: for HTTP errors
+           ConnectionError: If the request fails
+        """
+        url = urljoin(self.api_url + "/", path)
```
cosmogp: Wouldn't it be more helpful to make sure that the `api_url` ends with a `/` when assigned a new value?

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # NOTE: We only do this for backwards compatibility with the old API (only for local endpoint).
+        # The new endpoint is actually V1.
+        parsed_url = urlparse(self.api_url)
+        if self.is_v2 and parsed_url.scheme == "http" and not url.endswith("/"):
```
cosmogp, line 170: I think HTTPS might work as well? Am I wrong?

actions/src/sema4ai/actions/agent/__init__.py
cosmogp: Left it in another comment as well... maybe it would be simpler to have the `sema4_api_key` value be either a non-empty string or None - "LOCAL" seems out of place (this is taken from the `sema4ai.actions.agent.md`

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,286 @@
...
+        if len(parts) >= 4 and "conversations" in parts:
+            conv_index = parts.index("conversations") + 1
+            if conv_index < len(parts):
+                conversation_id_only = parts[conv_index]
+                log.info(f"Extracted conversation_id: {conversation_id_only} from path")
+            else:
+                # Invalid path format
+                raise AgentApiClientException(
+                    f"Invalid path format in conversation ID: {conversation_id}"
+                )
```
cosmogp, line 208: Isn't there a case where this might fail? Maybe raise an exception?
`if len(parts) < 4 or ...`

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,286 @@
...
+def send_message(
+    conversation_id: str, agent_id: str, message: str, sema4_api_key: str | None = None
+) -> str:
+    """Sends a message within a conversation and retrieves the agent's response.
+
+    Args:
+        conversation_id: The ID of the conversation
+        agent_id: The ID of the agent to send message to
+        message: The message content to send
+        sema4_api_key: The API key for the Sema4 API if running in cloud. Use LOCAL if in Studio or SDK!
```
cosmogp: Mentioning this again... Maybe LOCAL is not needed, only None would be sufficient.

cosmogp (commented):
Some comments, but LG

actions/src/sema4ai/actions/agent/_client.py
OvidiuCode: doesn't work, locally at least

actions/docs/api/sema4ai.actions.agent.md
OvidiuCode: we removed the "LOCAL" completely, need to adapt this

actions/docs/api/sema4ai.actions.agent.md
OvidiuCode: Good point

actions/src/sema4ai/actions/agent/__init__.py
OvidiuCode: yes, this is a leftover, forgot to update in this place

fabioz (changes requested):
This must work without any `sema4_api_key`...

The user doesn't have this. We should be able to get this from our environment in both cases (cloud and local).

Without this change this can't really go in...

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,286 @@
...
+        name: The name of the agent
+        sema4_api_key: The API key for the Sema4 API if running in cloud. Leave empty if in Studio or SDK!
+
+    Returns:
+        The agent that matches the given name.
+    """
+    # Initialize client with API key if provided
+    client = _AgentAPIClient(api_key=sema4_api_key)
+
+    agents = _get_agents(client, "agents")
```
fabioz, line 84: I believe the agent server has a proper `agents/by-name` API you can use...

fabioz:
One thing is that I think that the most useful API which is the new `prompt` API is not being covered... not that the current API being made available isn't useful, but the most useful API would probably be the new `prompt` API.

actions/src/sema4ai/actions/agent/__init__.py
OvidiuCode: this is only available for the newer versions of agent server and it does return all agents that start with that name, which is a little bit confusing if we expect to only get one. These things we can improve later on if there is a customer need or something. I originally made it with the endpoint but then switched back to filtering to keep the compatibility with what we have now in agent connector

OvidiuCode:
@fabioz from what I can see in the openapi spec is that the /prompt API is not yet available in the public API, only in private. We can integrate it later when it is available

OvidiuCode:
Probably won't be used anymore so closing this.

PR #246, opened 2025-07-03
Add agent module with prompt_generate function

OvidiuCode:


actions/src/sema4ai/actions/agent/_models.py
```diff
@@ -0,0 +1,569 @@
...
+
+PromptImageMimeType = Literal["image/jpeg", "image/png", "image/gif", "image/webp"]
+PROMPT_IMAGE_MIME_TYPES = {
+    "image/jpeg",
+    "image/png",
+    "image/gif",
+    "image/webp",
+}
+# TODO: Remove this once the files module is implemented
+UploadedFile = Any
```
fabioz: I think `UploadedFile` doesn't make sense for us in actions.

fabioz: Or maybe it does (I guess it's the case where the user uploads a file to the chat and references it later, but we need to have a proper API or remove it...)

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+    ReductoPlatformParameters,
+)
+
+log = logging.getLogger(__name__)
+
+
+def _get_id_from_action_context(action_context, id_key: str) -> str:
+    value = action_context.value
+    if not isinstance(value, dict):
+        raise RuntimeError("Action context value is not a dictionary!")
```
fabioz: I think we should try to use `ActionError` whenever we can (otherwise the user just sees an "Internal Error" if something goes wrong).

Note that in this case, the result goes up to the LLM, so, making it more descriptive is better (for instance `"Internal Error: Unable to get the Action Context because it's not a dictionary"`)

Likewise for the other errors raised in the changes...

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+        raise RuntimeError(f"No {id_key} found in the invocation_context!")
+
+    if not isinstance(id_value, str):
+        raise RuntimeError(
+            f"{id_key} is not a string (found: {id_value} ({type(id_value)}))!"
+        )
+    return id_value
+
+
+def _get_id_generic(id_key: str, header_key: str | None = None) -> str:
```
fabioz: Do we actually have any place that calls this with a `None` header key? If not I'd remove the `|None=None`

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+    if id_value is None:
+        # Ok, unable to get the id through any of the heuristics,
+        # let's raise an error.
+        raise RuntimeError(f"Unable to get the {id_key}")
+
+    return id_value
+
+
+def get_thread_id() -> str:
+    """Get the thread ID from the action context or the request headers."""
```
fabioz: This needs better docs to say that it'll raise an error if the thread id cannot be found (which is expected when running in VSCode unless a `x-invoked_for_thread_id` is passed as a header in the request in the configured input).

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+    return _get_id_generic("thread_id", "x-invoked_for_thread_id")
+
+
+def get_agent_id() -> str:
+    """Get the agent ID from the action context or the request headers."""
+    return _get_id_generic("agent_id", "x-invoked_for_agent_id")
```
fabioz: This needs better docs to say that it'll raise an error if the thread id cannot be found (which is expected when running in VSCode unless a `x-invoked_for_agent_id` is passed as a header in the request in the configured input).

actions/src/sema4ai/actions/agent/__init__.py
OvidiuCode: right, this is mandatory

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+@validate_call
+def prompt_generate(
+    prompt: Prompt,
```
fabioz: I think you should accept the `prompt` as a `dict` as an escape hatch if there's something we hadn't thought of so there's no validation in our side in that case.

Likewise for `platform_config`.

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,185 @@
...
+from urllib.parse import urlencode, urljoin, urlparse
+
+import sema4ai_http
+
+from sema4ai.actions._action import get_x_action_invocation_context
+
+log = logging.getLogger(__name__)
+
+
+class AgentApiClientException(Exception):
```
fabioz: Subclass `ActionError`.

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,185 @@
...
+            return False
+
+    def _get_pid_file_path(self) -> str:
+        """Get the path to the agent-server.pid file based on the operating system.
+
+        Returns:
+            str: Path to the PID file
+        """
+        # Determine OS-specific path
+        if platform.system() == "Windows":
```
fabioz: Minor: Always prefer `sys.platform == 'win32'`.

Reason: `platform.system()` actually calls `uname()` which will actually spawn a subprocess to collect the information. `sys.platform` is always just there as an attribute.

actions/src/sema4ai/actions/agent/_client.py
```diff
@@ -0,0 +1,185 @@
...
+        Returns:
+            sema4ai_http.ResponseWrapper object
+
+        Raises:
+           ValueError: for unsupported HTTP methods
+           AgentApiClientException: for HTTP errors
+           ConnectionError: If the request fails
+        """
+        url = self.api_url
+        if not url.endswith("/"):
```
fabioz: Now that you use `urljoin`, I don't think you need this condition anymore.

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,204 @@
...
+    platform_config: AnyPlatformParameters | None = None,
+    thread_id: str | None = None,
+    agent_id: str | None = None,
+):
```
fabioz: We should provide actual typing for the response...

actions/src/sema4ai/actions/agent/_client.py
OvidiuCode: good catch

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    return _get_id_generic("agent_id", "x-invoked_by_assistant_id")
+
+
+def prompt_generate(
+    prompt: Prompt | dict,
+    platform_config: AnyPlatformParameters | dict | None = None,
+    thread_id: str | None = None,
+    agent_id: str | None = None,
+) -> dict:
```
fabioz: Can we get the actual shape of the return in typing as a proper typeddict instead of a plain `dict`?

actions/src/sema4ai/actions/agent/_models.py
```diff
@@ -0,0 +1,522 @@
...
+        Field(
+            default=None,
+            description="Initial instruction that defines the AI's behavior and context",
+        ),
+    ] = None
+
+    messages: Annotated[
+        list[
+            Annotated[
+                Union[
```
fabioz: Minor: use `|` syntax instead of `Union`.

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -0,0 +1,234 @@
...
+    """Gives a prompt to an agent.
+
+    Args:
+        prompt: The prompt to generate.
+        platform_config: The platform configuration if method is called without action context (optional).
+        thread_id: The thread ID to use for the prompt (optional).
+        agent_id: The agent ID to use for the prompt (optional).
+
+    Note:
+        Either platform_config, thread_id or agent_id must be provided when calling this method without action context.
```
fabioz, line 180: Could also mention the actual reason (i.e.: the platform_config will be automatically gotten from the agent which is related to the thread_id or agent_id).

actions/src/sema4ai/actions/agent/_platforms.py
```diff
@@ -0,0 +1,378 @@
...
+    model_config = ConfigDict(extra="forbid")
+
+    kind: Literal["groq"] = Field(
+        default="groq",
+        description="The kind of platform parameters.",
+    )
+
+    groq_api_key: str | None = Field(
+        default=None,
+        description="The Groq API key. If not provided, it will be "
```
fabioz, line 324: I think "attempted to be inferred from the environment" -- here and in other places -- is too abstract (i.e.: how is it attempted to be inferred?)

OvidiuCode: That's deep in the agent server implementation. I think there are some platform specific keys. I could look and maybe provide the env var

fabioz (approved):

PR #271, opened 2025-09-09
If the `x-action-invocation-context` header is present, it should not…

fabioz:
… change how the arguments are processes in the MCP use case.

### DIFF
--- a/action_server/docs/CHANGELOG.md
+++ b/action_server/docs/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 2.14.3 - 2025-09-09
+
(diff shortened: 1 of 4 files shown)


sukeesh (approved):

action_server/docs/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- If the `x-action-invocation-context` header is present, it should not change how the arguments are processes in the MCP use case.
```
sukeesh:
```suggestion
- If the `x-action-invocation-context` header is present, it should not change how the arguments are processed in the MCP use case.
```
?

PR #273, opened 2025-09-24
Add the ability to specify llm model to prompt_generate function

OvidiuCode:


jjhenkel (approved):
Makes sense, this will be really nice to have so folks can pick and choose a precise model to use for certain scenarios (say, wanting to use a `gpt-5-thinking-high` to do a "hard task")

fabioz (approved):

PR #279, opened 2025-10-08
Add SecretSpec class for tagging secrets

OvidiuCode:
Tested in an action package:
- works like a Secret
- metadata generates the tag correctly in the object
- linting works as expected


actions/README.md
```diff
@@ -58,6 +58,81 @@ You can explicitly provide the `is_consequential` flag for an action to mark it'
 def get_weather_forecast(city: str, days: int, scale: str = "celsius") -> str:
 ```
 
+### Working with Secrets
```
fabioz, line 61: This should be in a custom doc in https://github.com/Sema4AI/actions/tree/master/action_server/docs/guides, probably just a new section in https://github.com/Sema4AI/actions/blob/master/action_server/docs/guides/07-secrets.md

I guess we should have some links to those guides from this readme too (maybe some auto-generated TOC when building the docs, although that's probably out of the scope of this PR).

actions/src/sema4ai/actions/_managed_parameters.py
```diff
@@ -85,6 +85,16 @@ def _is_secret_node(self, param_name: str, node: FunctionDef) -> bool:
...
                         "sema4ai.actions.Secret",
+                        "DocumentIntelligenceSecret",
+                    ):
+                        return True
+
+                    # Check for Annotated[Secret, ...] patterns
+                    # Handle: Annotated[Secret, SecretSpec(...)]
+                    if "Annotated[" in unparsed and (
+                        "Secret," in unparsed
+                        or "actions.Secret," in unparsed
```
fabioz: I think you should just check `Secret,` and remove the `actions.Secret` and `sema4ai.actions.Secret` as those will be already matched by the first one.

actions/src/sema4ai/actions/_secret/__init__.py
```diff
@@ -101,6 +115,39 @@ def value(self) -> str:
...
+        self.tag = tag
+
+
+def get_secret_spec_from_annotation(annotation: type) -> "SecretSpec | None":
+    """Extract SecretSpec from an Annotated type annotation if it exists."""
+    from typing import get_args, get_origin
+
+    try:
+        origin = get_origin(annotation)
+        if origin is not None and hasattr(typing, "Annotated"):
```
fabioz: I don't think you need the `hasattr(typing, "Annotated")`, just access it directly -- it should be always there on the versions of python we support. You don't even need to check if origin is None, just check if `origin is typing.Annotated`.

actions/src/sema4ai/actions/_secret/__init__.py
```diff
@@ -101,6 +115,39 @@ def value(self) -> str:
...
+            if origin is getattr(typing, "Annotated", None):
+                args = get_args(annotation)
+                # Check metadata for SecretSpec
+                for metadata in args[1:]:
+                    if isinstance(metadata, SecretSpec):
+                        return metadata
+    except Exception:
+        pass
```
fabioz: We don't want to fail silently if there's some error in the heuristic let it bubble up for us to fix it as it'd be a bug.

actions/src/sema4ai/actions/_secret/__init__.py
```diff
@@ -6,6 +6,20 @@
 
 
 def is_secret_subclass(cls: type) -> bool:
+    from typing import get_args, get_origin
+
+    # Handle Annotated types (e.g., Annotated[Secret, SecretSpec(...)])
+    try:
+        origin = get_origin(cls)
+        if origin is not None and hasattr(typing, "Annotated"):
```
fabioz: The exception/typing.Annotated comments also apply to this...

fabioz (changes requested):

fabioz (approved):

PR #288, opened 2025-10-17
SDM-41: Add Client APIs for Fetching Dataframes from Agent-Server

rajeshyum:
Improve the Table class in sema4ai-actions to better support dataframe workflows. This includes strengthening validation, adding metadata support, and enabling actions to fetch dataframes from the agent-server via client APIs.


actions/src/sema4ai/actions/_table.py
```diff
@@ -15,12 +15,22 @@ class Table(pydantic.BaseModel):
 
     columns: list[str]
     rows: list[Row]
-
-    def __init__(self, columns: list[str], rows: list[Row]):
+    name: str | None = None
```
fabioz: We should not break backward compatibility here, so, this means that if the user doesn't pass a `name` or a `description`, the `model_dump` must not have a `name` nor `description` if they're None. Can you create a test for that?

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -217,10 +221,123 @@ def prompt_generate(
...
+
+        return Table(
+            columns=data["columns"],
+            rows=data["rows"],
+            name=data.get("name"),
+            description=data.get("description"),
+        )
+    except AgentApiClientException as e:
+        # Convert 404 to ValueError for better UX
+        if "404" in str(e):
```
fabioz: I think you should improve the `AgentApiClientException` to actually have the error code so that you can do a check for the error code and not `404 in str(e)`.

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -217,10 +221,123 @@ def prompt_generate(
     return response.json()
 
 
+def list_data_frames() -> list[dict]:
```
fabioz: The `list[dict]` as the return is not a really good API, create a typed dict with the expected return so that it's a `list[DataFrameInfo]`

actions/tests/actions_tests/test_actions_custom_data_models/test_table_result.yml
```diff
@@ -3,6 +3,8 @@ result:
   columns:
   - a
   - b
+  description: null
```
fabioz, line 6: This actually means we broke backward compatibility because additional fields are available in the Table model_dump (so, we should update to 2.0 and note that in the changelog accordingly).

actions/tests/actions_tests/test_agent_dataframe_functions.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def test_list_data_frames_api():
+    """Test list_data_frames with mocked API client."""
+    from unittest.mock import MagicMock, patch
+
+    from sema4ai.actions.agent import list_data_frames
+
+    # Mock the _AgentAPIClient
+    with patch("sema4ai.actions.agent._AgentAPIClient") as MockClient:
```
fabioz: Note: the option not to use patching is using the dummy agent server -- search for the `agent_dummy_server` fixture.

Then the idea would be that instead of patching you'd implement the API expected in the dummy server as needed.

It's a bit better, but maybe not better enough to change all the tests now that they're already done this way though, so, you can leave as is ;)

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -217,10 +221,123 @@ def prompt_generate(
...
+            query_params={"thread_id": thread_id},
+        )
+        return response.json()
+    except AgentApiClientException as e:
+        raise ActionError(
+            f"Failed to list dataframes from thread {thread_id}: {e}"
+        ) from e
+
+
+def get_data_frame(name: str, limit: int = 10000) -> "Table":
```
fabioz: I think we should add support for more parameters (which are already accepted in the agent server):
- offset
- column_names
- order_by

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -217,10 +221,123 @@ def prompt_generate(
...
+    thread_id = get_thread_id()
+    client = _AgentAPIClient()
+
+    try:
+        response = client.request(
+            f"data-frames/{name}",
+            method="GET",
+            query_params={"thread_id": thread_id, "limit": limit},
+        )
+        data = response.json()
```
fabioz: This flow is reasonable, but ideally we'd check if:
- Is `pyarrow` available (importable)? -- cache that information.
- If it is, request the data in parquet format and then create the table from the parquet contents (and then read the parquet contents and create the table from it).
- if `pyarrow` isn't available, request in json format as it is now.

As a note, for future work -- not this PR -- we should provide new APIs to get the data frame in different formats (numpy/pandas) in which case we'd actually only get the data in the `parquet` format instead of json (we shouldn't make those required dependencies, so, it'd be fine to fail in those if they're not there).

We need to be careful not to import those globally anywhere though (as they should only be optional dependencies).

fabioz (changes requested):

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -8,9 +8,75 @@
...
+                rows.append(row)
+            
+            return {
+                "columns": columns,
+                "rows": rows,
+                "name": None,  # Parquet doesn't include metadata
+                "description": None
+            }
+        except Exception:
+            # Parquet parsing failed, fall back to JSON
```
fabioz: I think you should just let the exception be raised here (if we requested in parquet format, it'd be surprising to have the response as json).

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -307,12 +390,31 @@ def get_data_frame(name: str, limit: int = 10000) -> "Table":
...
         response = client.request(
             f"data-frames/{name}",
             method="GET",
-            query_params={"thread_id": thread_id, "limit": limit},
+            query_params=query_params,
         )
-        data = response.json()
+        
+        # Try to parse as Parquet first, fallback to JSON
+        data = _parse_dataframe_response(response)
```
fabioz: You know the format requested (from a few lines up), so, you probably want to have a `_parse_dataframe_response_from_parquet` if parquet was requested and otherwise just call `response.json()`.

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -8,9 +8,75 @@
...
+        response: HTTP response object
+        
+    Returns:
+        dict: Parsed dataframe data with columns, rows, name, description
+    """
+    import json
+    
+    # Check if response is Parquet format and PyArrow is available
+    content_type = response.headers.get('content-type', '').lower()
+    if ('parquet' in content_type or 'application/octet-stream' in content_type) and _is_pyarrow_available():
```
fabioz: It should not need to check the content type as you should know whether the request was for parquet or json already.

actions/src/sema4ai/actions/agent/__init__.py
rajeshyum: Enhanced error handling with status_code

actions/tests/actions_tests/test_actions_custom_data_models/test_table_result.yml
rajeshyum: Updated version to 1.6.0 since sema4ai-data requires sema4ai-actions >=1.2.0,<2.0.0. and its defined inside the poetry.lock

actions/src/sema4ai/actions/agent/__init__.py
rajeshyum: Updated to support parameters (offset, column_names, order_by)

actions/src/sema4ai/actions/agent/__init__.py
rajeshyum: Updated to add PyArrow/Parquet Support - With JSON fallback

actions/src/sema4ai/actions/agent/__init__.py
rajeshyum: added _parse_dataframe_response_from_parquet() function for parquet request

fabioz:
@codex review

actions/src/sema4ai/actions/agent/__init__.py
rajeshyum: removed the check

actions/src/sema4ai/actions/agent/__init__.py
rajeshyum: updated to list[DataFrameInfo]

actions/tests/actions_tests/test_agent_dataframe_functions.py
rajeshyum: Updated to use agent_dummy_server instead of patch

actions/src/sema4ai/actions/agent/__init__.py
```diff
@@ -217,10 +277,156 @@ def prompt_generate(
...
+        return Table(
+            columns=data["columns"],
+            rows=data["rows"],
+            name=data.get("name"),
+            description=data.get("description"),
+        )
+    except AgentApiClientException as e:
+        # Convert 404 to ValueError for better UX
+        if e.status_code == 404:
+            raise ValueError(
```
fabioz: I think it should be an `ActionError` (if it's a `ValueError`, the message is not shown to the client, only `ActionError` messages are considered "safe" to be shown).

rajeshyum: Updated to ActionError instead of valueError

fabioz (approved):

PR #289, opened 2025-10-22
Add `list_files` to `chat` module

mikahanninen:
This would help list all files that exists for the chat thread.

Returns a list of filenames.


fabioz (approved):

fabioz (commented):
Just missing an entry in the changelog.

PR #290, opened 2025-10-24
Add json_to_dataframe utility function

OvidiuCode:


OvidiuCode:
We will most likely go with Josh jq solution, so closing this.

PR #298, opened 2025-11-07
Fix get and list data frames

OvidiuCode:


fabioz:
Seems like it needs an update to the docs.

Also, while you're at it, can you take a look at the docstrings, it seems the current docstrings break the generated docs: https://github.com/Sema4AI/actions/blob/master/actions/docs/api/sema4ai.actions.agent.md#list_data_frames

fabioz (approved):

fabioz:
Still needs to pass tests, but the changes seem fine ;)

PR #304, opened 2025-11-26
Fix issue (websockets api connect expects `additional_headers` instea…

fabioz:
…d of `extra_headers` in action-server --expose.


sahokas (approved):
Thanks Fabio, seems straightforward and also tested to be working.

Note: --expose still does not work with MCP using streamable-http, but that is effectively a missing feature in the reverse proxy

PR #306, opened 2025-12-03
DEV-2387 Separate file name validations on windows vs. unix

joshelser:
Windows disallows certain characters in file names that Unix does not. This is required because the AgentServer uses the local filesystem for file storage in Studio. However, we should not over-enforce this limit as it creates a situation where ACE/SPAR/etc then cannot store files with names that the OS does allow.

This change separates the `_check_that_name_is_valid_in_filesystem` function into unix and windows equivalent functions, delegating to the appropriate one depending on platform.

### DIFF
--- a/actions/src/sema4ai/actions/chat/__init__.py
+++ b/actions/src/sema4ai/actions/chat/__init__.py
@@ -1,6 +1,6 @@
 import logging
 import os
-import platform
+import sys
 import typing
 from functools import lru_cache
 from pathlib import Path
@@ -80,7 +80,7 @@ def _check_that_name_is_valid_in_filesystem(name: str):
     if not name or name in (".", ".."):
         raise ValueError(f"Name '{name}' is not valid.")
 
-    is_windows = platform.system() == "Windows"
+    is_windows = sys.platform == "win32"
 
     if is_windows:
         _check_windows_filename(name)
--- a/actions/src/sema4ai/actions/chat/__init__.py
+++ b/actions/src/sema4ai/actions/chat/__init__.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import platform
 import typing
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


actions/src/sema4ai/actions/chat/__init__.py
```diff
@@ -48,12 +55,37 @@ def _check_that_name_is_valid_in_filesystem(name: str):
...
+        raise ValueError(f"Name '{name}' cannot contain null byte.")
+
+
+def _check_that_name_is_valid_in_filesystem(name: str):
+    """Validate that a name is safe for filesystem use on the current platform."""
+    # Common checks: empty name or relative path indicators
+    if not name or name in (".", ".."):
+        raise ValueError(f"Name '{name}' is not valid.")
+
+    is_windows = platform.system() == "Windows"
```
fabioz: Minor, but always prefer `sys.platform == "win32"` -- platform.system() is much more heavywheight, must do a subprocess call and is less stable (should only be used for diagnosis, not really for checking the platform).

See: https://discuss.python.org/t/clarify-usage-of-platform-system/70900 for more info.

joshelser: oh, thanks. I had no idea which one was preferred. I will switch.

actions/tests/actions_tests/test_chat_files.py
```diff
@@ -229,3 +230,241 @@ def test_actions_with_agent_headers_call_with_invocation_context(datadir: Path):
     # assert json_output.read_text() == "value-in-header"
     output = result.stdout.decode("utf-8")
     assert "thread_id: thread-id-123" in output
+
+
+class TestFilenameValidationWindows:
+    """Test filename validation on Windows platform."""
+
+    @patch("platform.system", return_value="Windows")
+    def test_valid_names(self, mock_platform):
```
fabioz: Why is `mock_platform` here (and in other tests)?

joshelser: cursor slop I didn't catch -- removed.

actions/src/sema4ai/actions/chat/__init__.py
```diff
@@ -48,12 +55,37 @@ def _check_that_name_is_valid_in_filesystem(name: str):
...
+        raise ValueError(f"Name '{name}' is a reserved Windows device name.")
 
-    # Filenames cannot end in a space or dot.
+    # Cannot end in space or dot on Windows
     if name.endswith(" ") or name.endswith("."):
-        raise ValueError(f"Name {name} cannot end in a space or dot.")
+        raise ValueError(f"Name '{name}' cannot end in a space or dot on Windows.")
+
+
+def _check_unix_filename(name: str):
```
kariharju, line 67: Even though these are the only invalid chars, I think it is still recommended to avoid control chars etc.
So potentially I would do: `if any(ord(char) < 32 for char in name):` to all.

Also thinking that can the different rules lead to confusion in cases where we have people using chat files from all platforms and sharing them with each other

Thinking:
- “native mode” = minimum illegal chars for each platform
- “portable mode” = intersection of safe rules across platforms/tools

But as is this is good-to-go as it tackles the cases I think clearer, we can do other rounds.
(still amazes me that these are not in the core of every platform that claims cross-platform)

kariharju (approved):

fabioz (approved):

actions/src/sema4ai/actions/chat/__init__.py
joshelser: Yeah, I think I have another PR in my future to further restrict the edge-cases of file names. At least, it is significantly less likely to have a filename with non-ascii range characters in them (compared to a ":").

PR #308, opened 2025-12-04
CVE updates for Action Server v2.17.1

kariharju:
CVE and CodeQL updates on Action Server
Action server v2.17.1

CodeQL raised the point that blocking TLSv1x is a high level finding.
- Added the checks: https://docs.python.org/3/library/ssl.html#ssl.OP_NO_TLSv1


action_server/src/sema4ai/action_server/vendored_deps/url_callback_server.py
```diff
@@ -106,6 +106,13 @@ def _wrap_socket(
         raise ValueError("certfile must be specified")
     context = SSLContext(ssl.PROTOCOL_TLS_SERVER)
     context.verify_mode = ssl.VerifyMode.CERT_NONE
+    # Ensure only TLSv1.2 and newer are accepted
```
fabioz, line 109: I'll take your word for it ;)

fabioz (approved):

action_server/src/sema4ai/action_server/vendored_deps/url_callback_server.py
kariharju: This is the way it should go and no one should be using TLSv1x anywhere anymore.

CodeQL, Claude and Cursor all agreed on the python code here 😅
The flags are specced here:
https://docs.python.org/3/library/ssl.html#ssl.OP_NO_TLSv1

..but waiting for the tests to pass.

PR #320, opened 2026-01-08
Drop mac intel from build / Release Action Server 3.0.0

fabioz:


action_server/docs/CHANGELOG.md
```diff
@@ -2,10 +2,15 @@
 
 ## Unreleased
 
+## 2.17.4 - 2026-01-08
```
kariharju: I would clearly state this as a major version update.. at least a minor one.

kariharju (approved):
LGTM, the version number bump is something that I would think about.
If some poor soul is running on an old Intel mac these changes will hit like a brick wall.

OvidiuCode (approved):

